PR #43, opened 2024-03-12
Make client_config `pub`

0xmovses:
I need access to type `WalletKey` when using `EthersClientConfig` in my tests, but I'm unable to because `client_config` is private. 

I tried to get around this by creating by `WalletKey` and subsequently `EthersClientConfig` types, but I hit a snag when calling the relayer like so:
```rust
relayer.run(ethers_client_config.clone()
```

I get the error:
```
mismatched types
`eth_client::EthersClientConfig` and `EthersClientConfig` have similar names, but are actually distinct typesrustcClick for full compiler diagnostic
publisher_tests.rs(72, 26): arguments to this method are incorrect
lib.rs(54, 1): `eth_client::EthersClientConfig` is defined in crate `eth_client`
client_config.rs(66, 1): `EthersClientConfig` is defined in crate `risc0_ethereum_relay`
lib.rs(71, 18): method defined here
```
I need to use the actual type from this repo. Is there any downside to making this module public? If not it's difficult to write test code for the `Relayer`.

### DIFF
--- a/relay/src/lib.rs
+++ b/relay/src/lib.rs
@@ -15,7 +15,7 @@
 #![doc = include_str!("../README.md")]
 
 mod api;
-mod client_config;
+pub mod client_config;
 mod downloader;
 pub mod sdk;
 mod storage;


lib/openzeppelin-contracts
capossele, line 1: any particular reason to update this library?

lib/solidity-bytes-utils
capossele, line 1: Same question as above

0xmovses: Unintended change. Will remove it.

capossele (approved):
Thanks!

PR #50, opened 2024-03-21
Bootstrap Poseidon2 changes

tzerrell:
Bootstrap the changes from https://github.com/risc0/risc0/pull/1571 and https://github.com/risc0/risc0/pull/1626


contracts/src/groth16/ControlID.sol
```diff
@@ -20,6 +20,7 @@
 pragma solidity ^0.8.9;
 
 library ControlID {
-    uint256 public constant CONTROL_ID_0 = 0x39ff805954f4eb2868d338764408f76d;
-    uint256 public constant CONTROL_ID_1 = 0x15cf3a5f4097269e3a6d921c18625531;
+    uint256 public constant CONTROL_ID_0 = 0x17da5b0946fcb1115de33dca38ca0d4c;
+    uint256 public constant CONTROL_ID_1 = 0x4fa8b5b0563340e93606acfe2f6ab8d5;
```
nategraf, line 24: This is a standing issue, but we need to clarify these names. In particular, `CONTROL_ID_x` is mileading because there are actually the control root / allowed program root.

tzerrell: Yeah, I need to open an issue, I had a discussion w/ Paul/Frank/Jeremy choosing the name `BN254_CONTROL_ID` where we discussed also cleaning up the other related names (including these) -- but it was not on ceremony critical path so I haven't done it yet.

capossele (approved):

nategraf:
We will not want to merge this unless we additionally change the `Cargo.toml` files to reference a version of `risc0` that includes these changes. Additionally, with https://github.com/risc0/risc0/pull/1623 in the pipeline, it seems the next release of `risc0` will include those circuit changes as well. As such, will want to update this PR to include those (and rename it to reflect the total set of circuit changes included)

capossele:
Ah yeah you’re right

capossele:
Marked as draft until then

nategraf:
Was Poseidon2 rolled in https://github.com/risc0/risc0-ethereum/pull/68? If so, should this PR be closed?

nategraf:
Changes here were rolled into subsequent PRs including bootstrapping.

tzerrell:
Thanks for closing!

PR #60, opened 2024-03-26
[Feat] View Call Proofs

capossele:
- add view call proofs library
- add erc20 example
- add erc20-counter examples

We should add here on in a different PR some test that can easily run on CI for both of the above examples
Once this land, we should also update the risc0 repo to add this library in the risc0-ethereum CI local patch


Wollac (approved):
lgtm

Issue #30, opened 2024-03-05
Fix documentation on the `CONTROL_ID` constants in the `RiscZeroGroth16Verifier.sol`


nategraf:
Comment highlighted below is inaccurate:
https://github.com/risc0/risc0-ethereum/blob/37e4cb9805f9180364a66cdef919d7c32e6fe8c8/contracts/src/groth16/RiscZeroGroth16Verifier.sol#L89-L95

pdg744:
Also not clear to me what `control id` is being referenced here: https://github.com/risc0/risc0-ethereum/blob/main/contracts/src/groth16/ControlID.sol

nategraf:
Addressing this in a pair of PRs. https://github.com/risc0/risc0-ethereum/pull/80 to `risc0-ethereum` is the first, and I'll open one to `risc0` with the associated bootstrap changes.

PR #80, opened 2024-04-24
Add emergency stop and verifier router contracts for version management

nategraf:
As we get ready to deploy RISC Zero verifiers to Ethewreum mainnet, we need a system to manage versions of the verifier contract and proof system. In particular, we need a way to respond to critical bugs, and we need a way to make upgraded versions of the proof system available to developers. Both mechanisms need to be opt-in so that applications can choose the solution that works best for their security model.

This PR implements a basic version management system, based around two new contracts:

* `RiscZeroVerifierEmergencyStop` is a wrapper around a verifier contract that allows for the verifier to be shutdown if a critical bug is discovered.
* `RiscZeroVerifierRouter` is a wrapper around a list of verifiers that allows new verifiers to be added to support proof system upgrades.

These are intended to be used together with an authorized multisig, and a `TimelockController` on addition of new verifiers.

This PR is currently marked as draft, as the verifier identifiers are not yet fully implemented in the `risc0` codebase.

Also addresses https://github.com/risc0/risc0-ethereum/issues/30 as part of a refactor to the `IRiscZeroVerifier` interface.

Related to https://github.com/risc0/risc0/pull/1794, from which the verifier parameters have been bootstrapped.


nategraf:
@capossele, this is ready for review when you have a chance

contracts/src/RiscZeroVerifierRouter.sol
```diff
@@ -0,0 +1,103 @@
...
+        if (verifiers[selector] == TOMBSTONE) {
+            revert SelectorRemoved({selector: selector});
+        }
+        if (verifiers[selector] != UNSET) {
+            revert SelectorInUse({selector: selector});
+        }
+        verifiers[selector] = verifier;
+    }
+
+    /// @notice Removes verifier from the router, such that it can no receive verification calls.
```
pdg744, line 59:
```suggestion
    /// @notice Removes verifier from the router, such that it can not receive verification calls.
```

contracts/src/RiscZeroVerifierRouter.sol
```diff
@@ -0,0 +1,103 @@
...
+    ///         Removing a selector sets it to the tombstone value. It can never be set to any
+    ///         other value, and can never be reused for a new verifier, in order to enfoce the
```
pdg744, line 61:
```suggestion
    ///         other value, and can never be reused for a new verifier, in order to enforce the
```

capossele (approved):
Awesome man!

PR #81, opened 2024-04-25
Give view call library a new name, Steel

nategraf:
Highlighting the power of this library, and it's goal of making Alloy available for provable (i.e. "hardened") computation in the zkVM, we are naming the library Steel.


pdg744:
Did we land on `Steel` or `STEEL`?

nategraf:
> Did we land on `Steel` or `STEEL`?

We landed on "Steel" as the official name.

capossele (approved):

nategraf:
NOTE: The `examples` job is currently failing because the `main` branch here and the Prover deployment to Bonsai have incompatible control IDs / Groth16 setup. We plan on addressing this fragile CI issue with an update to produce proofs locally in CI.

PR #91, opened 2024-04-30
feat: efficient multi view call support

### LINKED ISSUE #90: Improve handling of the view call env for multiple view calls
Feedback from @austinabell in his usage of the Steel view call library.

When making multiple view calls within a single guest execution with Steel, there is a problem that the [ViewCallEnv](https://docs.rs/risc0-steel/latest/risc0_steel/struct.ViewCallEnv.html) cannot be cloned, but the [execute method](https://docs.rs/risc0-steel/latest/risc0_steel/struct.ViewCall.html#method.execute) consumes the object.

austinabell:
Proposal to close #90 

The least changing API would be to switch the `preflight` and `execute` methods to take in a mutable reference to the `env`, but this comes with the downside of having a breaking change from previous APIs (unless using different names) and an awkward API imo.

This simple change mentioned above would look sth like:

```rust
 	// Host
    let mut env = EthViewCallEnv::from_provider(provider, BLOCK)?;
    let _result = ViewCall::new(call, contract).preflight(&mut env)?;
    let _result2 = ViewCall::new(call2, contract).preflight(&mut env)?;
    let input = env.into_zkvm_input()?

	// Guest
    let env = input.into_env();
    let result = ViewCall::new(call, contract).execute(&env);
 	let result2 = ViewCall::new(call2, contract).execute(&env);
```

Where the proposal in this API flips the env and `ViewCall`, as it's more clear the call is operating on the `env`:

```rust
 	// Host
    let env = EthViewCallEnv::from_provider(provider, BLOCK)?;
    let _result = env.preflight(ViewCall::new(call, contract))?;
	let _result2 = env.preflight(ViewCall::new(call2, contract))?;
    let input = env.into_zkvm_input()?

	// Guest
    let env = input.into_env();
    let result = env.execute(ViewCall::new(call, contract));
 	let result2 = env.execute(ViewCall::new(call2, contract));
```

~~Opening as draft until approval, and I will add tests and an example with this.~~ basic test added and example could be #95, although that example probably isn't practical without historical queries supported better, in the example or in steel

Another option is to create APIs around a type like `MultiViewCall`, and describe the multiple calls that way which returns a tuple of results with the zkvm input. I didn't love this option, as it would just be confusing to use, but figured I'd mention that also. Happy to elaborate on this if helpful, though!


austinabell:
don't review yet. As discussed with Wolfgang, going to scope out the proposed alloy provider API to try to more closely match that if possible. Those APIs are unstable and not released yet, so it might not be a straightforward pick.

austinabell:
I'd generally recommend against trying to match their APIs until released/stabilized. There is too much unfinished that making the migration for things like the `Provider` APIs or to use the upstream `TransactionRequest` type, as those would just be less ergonomic. Some other notes:

- Provider would need to be async to be drop-in replacement, unclear if we want have asymmetric APIs from host/guest
- `call` APIs of their provider use a call specific block height, we might want to keep it provider specific to nudge developers to making requests at the same block height. Hash linking automatically between any range probably isn't ideal (maybe nice ergonomics, though)
- Should we use client-like naming from `env` to move towards being closer to alloy semantics? Also, perhaps `execute` should eventually be `call`, as there aren't equivalent semantics in alloy providers?

I don't have opinions about what the API looks like in this PR, but it intuitively seems worth it to make the change of these functions being impls of the `env`/`provider` as this is closer to alloy and makes a slightly cleaner API for consumption with supporting multi calls. Happy to switch to the other proposed in the PR description, though!

steel/testdata/rpc_cache.json
```diff
@@ -475,6 +475,76 @@
         ]
       }
     ],
+    [
+      {
+        "block_no": 19493153,
+        "address": "0xdac17f958d2ee523a2206206994597c13d831ec7",
```
Wollac, line 481: The same `storage_proof` is already in the cache, only with one index missing. This old proof should be removed to keep the cache as small as possible

Wollac (changes requested):
These changes make a lot of sense, especially in the case of the ERC20 example where multiple balances are queried.
I think the API of having to create a `ViewCall` which is then consumed by the `execute` is a bit unintuitive. But, honestly, I don't have a better idea, and it might just be the cleanest design...
The actual implementation looks very good to me though.

steel/testdata/rpc_cache.json
austinabell: Is there a clean way to prune the old data or does this have to be done manually or by regenerating the cache?

austinabell: Wait, what is the duplicate? I didn't remove any other test, so the query with one less value is also needed for the old test. Am I overlooking one here/did I remove the wrong one? Seems the queries have to match exactly.

Wollac:
> are you suggesting to switch to the previous pattern (first example in the PR description)?

No, sorry, I didn't explain it properly: I really like the design of using `env` type methods. I was just wondering if there might be some nicer designs for the `execute(ViewCall::new(...`. Maybe something like:
- `env.call(view_call, to).from(sender).execute()` and `env.call(view_call, to).from(sender).preflight()` so that the `.call` would give you some kind of builder. (I'm not sure it would work exactly like that, but maybe similar).
- Or maybe rename `ViewCall` to `CallBuilder` or something like that to make it clearer that the stuff inside the execute is supposed to be a builder.
- Or something else.
I just felt like we were using a `Builder` pattern here and it might make sense to make that a little clearer.
But if it is just me who has this feeling, I am perfectly fine with this design.

austinabell:
> * `env.call(view_call, to).from(sender).execute()` and `env.call(view_call, to).from(sender).preflight()` so that the `.call` would give you some kind of builder. (I'm not sure it would work exactly like that, but maybe similar).

Ah, I see, yeah I was initially avoiding this to try to keep things as close to alloy APIs. They don't have this type of builder API and they don't accept the type that would be added in `view_call`, so might be moving more away from what those might end up as.

> * Or maybe rename `ViewCall` to `CallBuilder` or something like that to make it clearer that the stuff inside the execute is supposed to be a builder.

Very possible, maybe that's a better name, but creates some breakages for something that will likely change in the future?

> * Or something else.
>   I just felt like we were using a `Builder` pattern here and it might make sense to make that a little clearer.
>   But if it is just me who has this feeling, I am perfectly fine with this design.

I agree it's not so ideal, and happy to change to whichever you think is best, but nothing seems like a silver bullet here. Maybe will be more clear with more usage?

steel/testdata/rpc_cache.json
Wollac: Oh, sorry, you are absolutely right!
I just created the test cache from scratch, and it came up with a file where several lines were deleted.  However, upon closer inspection, it also adds the same number of lines in a different location. I am a bit confused though, I was expecting the cache file to be deterministic 🤔

Wollac:
> Ah, I see, yeah I was initially avoiding this to try to keep things as close to alloy APIs. They don't have this type of builder API and they don't accept the type that would be added in `view_call`, so might be moving more away from what those might end up as.

Yeah, I see what you mean! However, in `alloy` you do 
```
instance.FUNCTION_NAME(...).from(...).gas(...).send()
```
so in some sense it would be a similar pattern, but of course not a compatible API as you mentioned. So I am not sure, if it would make sense.

Wollac (approved):

austinabell:
> > Ah, I see, yeah I was initially avoiding this to try to keep things as close to alloy APIs. They don't have this type of builder API and they don't accept the type that would be added in `view_call`, so might be moving more away from what those might end up as.
> 
> Yeah, I see what you mean! However, in `alloy` you do
> 
> ```
> instance.FUNCTION_NAME(...).from(...).gas(...).send()
> ```
> 
> so in some sense it would be a similar pattern, but of course not a compatible API as you mentioned. So I am not sure, if it would make sense.

oh, I must have overlooked those APIs, let me try to find and revisit to make sure. I was looking primarily at the alloy-providers API, and thought they hadn't wired that through yet. (if you have a link that would be helpful)

austinabell:
> ```
> instance.FUNCTION_NAME(...).from(...).gas(...).send()
> ```
> 
> so in some sense it would be a similar pattern, but of course not a compatible API as you mentioned. So I am not sure, if it would make sense.

Scoping this out, it seems we wouldn't be able to do the same pattern, because this relies on some codegen in the sol macro (generated from `#[sol(rpc)]`), unless we are open to forking and maintaining an alternative. Maybe we could create a wrapper that parses this just to get the function names, but it doesn't turn out cleanly.

Alternatively, we could do something like:

```rust
FUNCTION_NAMECall {..}.with_{env/provider}(env).from(..).gas(..).call()
```

Although that still doesn't match. Also going to explore one other thing, going to see if there is a somewhat hacky way we can intercept the provider calls before completely ruling out :)

Wollac:
Wow, thanks for all the research. I really like the `call_builder` idea! But it seems like it's a pretty big change for this PR, especially if we want to have a nice distinction between preflight and guest calls. So let's do this in another PR.

PR #95, opened 2024-05-01
Add compound APR example that uses multi-call

austinabell:
Not done with this yet, I would like to potentially expand this to be able to prove an average APR over multiple points in time, but I might separate that into another PR if we want to keep this example simple.

Opening as draft because it depends on #91, and it's unclear we want that or this example here. Just opening for visibility.


nategraf:
Superseded by #115

austinabell:
> Superseded by #115

I mean, it's pretty independent. This is just if we want an example that actually needs to call multiple functions. (will needed to be updated for new APIs though if that comes in)

Wollac:
This is a good example to show how multiple functions can be called with Steel. Let's use it.

examples/token-stats/README.md
```diff
@@ -0,0 +1,48 @@
+# Compound Token Stats (APR Proof)
+
+**An example that calls multiple view function on the Compound USDC contract to compute the APR.**
```
nategraf, line 3:
```suggestion
**An example that calls multiple view functions on the Compound USDC contract to compute the APR.**
```

examples/token-stats/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Prerequisites
+
+To get started, you need to have Rust installed. If you haven't done so, follow the instructions [here][install-rust].
+
+Next, install the `cargo risczero` tool. We'll use `cargo binstall` to facilitate this. Detailed instructions can be found at [cargo-binstall].
+
+```bash
+cargo install cargo-binstall
+cargo binstall cargo-risczero
```
nategraf, line 15: Some trickiness with this is that, on main, they will end up using two different versions of the zkVM. One is specified in the Cargo.toml here as the `risc0-zkvm` dependency and will be `main`, and the other is the latest release installed by `cargo binstall`. It's unclear whether we are guaranteeing this will work or not, post 1.0.

At the very least, the result right now is
```
called `Result::unwrap()` on an `Err` value: incompatible client version: 1.1.0-alpha.1, server version: 1.0.1
```

This issue is not unique to this example, and we are trying to deal with this on `risc0` as well.

nategraf (approved):
This is a well-factored example 👌

Can we go ahead and add this to the `examples` CI job? We should at least make sure it builds. Better if we can run it too. Avoiding issues due to external API calls may mean using cached test data

https://github.com/risc0/risc0-ethereum/blob/da65e227450ac6ae8b62e956d0899e44f4ac7614/.github/workflows/main.yml#L157-L159

examples/token-stats/README.md
flaub: Might be better to just link to the install page on the docs so that if we make changes to the installation (like moving to `rzup`), we don't need to update all the different descriptions of installation.

examples/token-stats/methods/guest/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[workspace]
+
+[dependencies]
+alloy-sol-types = { version = "0.7" }
+core = { path = "../../core" }
+risc0-steel = { path = "../../../../steel" }
+risc0-zkvm = { git = "https://github.com/risc0/risc0", branch = "main", default-features = false, features = ["std"] }
```
flaub, line 12: It'd be better to pin to a specific commit rather than using the main branch. It's very easy for this to break in the future.

flaub:
Generally we should try to avoid using `branch = main` when creating risc0 dependencies. Now that we have 1.0 out, is there any reason to use git dependencies?

PR #98, opened 2024-05-02
Add banner image to Steel README

nategraf:
- add comment on splitDigest
- run fmt
- add banner image to Steel readme


nategraf:
Oops, included extra changes

PR #105, opened 2024-05-10
Drop post state digest from `IRiscZeroVerifier.verify`

nategraf:
With https://github.com/risc0/risc0/pull/1807, the ending system state is constant for executions ending with exit code `Halted(_)`. We take advantage of this to provide a fixed value for the post state digest, and drop that argument from the common interfaces.


contracts/src/IRiscZeroVerifier.sol
```diff
@@ -49,26 +51,23 @@ struct ReceiptClaim {
...
-        returns (ReceiptClaim memory)
-    {
+    /// @dev Input hash and postStateDigest are set to all-zeros (i.e. no committed input, or
+    ///      final memory state), the exit code is (Halted, 0), and there are no assumptions
+    ///      (i.e. the receipt is unconditional).
+    function from(bytes32 imageId, bytes32 journalDigest) internal pure returns (ReceiptClaim memory) {
         return ReceiptClaim(
             imageId,
-            postStateDigest,
+            SystemState(0, bytes32(0)).digest(),
```
capossele, line 70: I wonder wouldn’t a constant be more efficient here?

capossele (approved):

contracts/src/IRiscZeroVerifier.sol
nategraf: Good idea. I probably shouldn't trust the Solidity compiler to automatically turn that into a constant.

nategraf: Addressed in https://github.com/risc0/risc0-ethereum/pull/105/commits/b5d9e7dd0fd068aebcc727ff5a3f13b7555073ea

PR #110, opened 2024-05-14
Use default_prover

### LINKED ISSUE #77: Use Anvil for view call library examples
Foundry team fixed the bugs in Anvil that were previously preventing usage of Anvil for local development with the view call library. We can now use it, and should update instructions and scripts.

capossele:
- Refactor examples and ffi to use the default_prover rather than bonsai-sdk
- Refactor erc20-counter example to use anvil instead of Ganache

Closes #77


contracts/src/groth16.rs
```diff
@@ -12,74 +12,27 @@
...
-    ///
-    /// [RiscZeroGroth16Verifier.sol]: https://github.com/risc0/risc0/blob/v0.20.1/bonsai/ethereum/contracts/groth16/RiscZeroGroth16Verifier.sol#L76-L81
-    #[derive(Debug)]
-    struct Seal {
-        uint256[2] a;
-        uint256[2][2] b;
-        uint256[2] c;
-    }
-}
+pub struct Seal {}
```
nategraf, line 19: Why have this type? It seems to simply be a container for the ABI encoding methods, and with that I think we might as well drop the struct and simply have the methods live on the module.

contracts/src/test/RiscZeroCheats.sol
```diff
@@ -47,7 +47,7 @@ abstract contract RiscZeroCheats is CommonBase {
     ///       Uses the Bonsai proving service to run the guest and produce an on-chain verifiable
     ///       SNARK attesting to the correctness of the journal output. URL and API key for Bonsai
     ///       should be specified using the BONSAI_API_URL and BONSAI_API_KEY environment variables.
-    function prove(string memory elf_path, bytes memory input) internal returns (bytes memory, bytes32, bytes memory) {
+    function prove(string memory elf_path, bytes memory input) internal returns (bytes memory, bytes memory) {
```
nategraf, line 50: Looks like the comment above is out of date now, with no post state digest.

examples/erc20-counter/deployment-guide.md
```diff
@@ -13,28 +13,21 @@ You can either:
...
+    anvil
     ```
 
-    Once ganache is started, look at its logs and copy any of the Private Keys. You'll need one for the next step.
-    Then, keep it running in the terminal, and switch to a new terminal.
+   Once anvil is started, keep it running in the terminal, and switch to a new terminal.
 
 2. Set your environment variables:
     > ***Note:*** *This requires having access to a Bonsai API Key. To request an API key [complete the form here](https://bonsai.xyz/apply).*
+    > Alternatively you can generate your proofs locally, assuming you have a machine with an x86 architecture and [Docker] installed. In this case do not export Bonsai related env variables.
```
nategraf, line 26: Suggesting a line-break here:
```suggestion
    >
    > Alternatively you can generate your proofs locally, assuming you have a machine with an x86 architecture and [Docker] installed. In this case do not export Bonsai related env variables.
```

examples/erc20-counter/deployment-guide.md
```diff
@@ -133,6 +126,7 @@ You can deploy the Counter contract on a testnet such as `Sepolia` and run an en
 
 1. Get access to Bonsai and an Ethereum node running on a given testnet, e.g., Sepolia (in this example, we will be using [Alchemy](https://www.alchemy.com/) as our Ethereum node provider) and export the following environment variables:
     > ***Note:*** *This requires having access to a Bonsai API Key. To request an API key [complete the form here](https://bonsai.xyz/apply).*
+    > Alternatively you can generate your proofs locally, assuming you have a machine with an x86 architecture and [Docker] installed. In this case do not export Bonsai related env variables.
```
nategraf, line 129:
```suggestion
    >
    > Alternatively you can generate your proofs locally, assuming you have a machine with an x86 architecture and [Docker] installed. In this case do not export Bonsai related env variables.
```

nategraf (approved):
Looks good! This is a very helpful refactor

PR #115, opened 2024-05-16
Implement call builder pattern to more closely match alloy APIs

austinabell:
More context here https://github.com/risc0/risc0-ethereum/pull/91#issuecomment-2100548707

This is built on #110 

Draft until:
- [x] Update examples and usages to new API
- [x] Add docs
- [x] Deprecate `ViewCall` constructor
- [x] Update old deprecation warning to point to this API
- [x] Remove function added in pull 91, as that was between released versions (keep if this comes in after release)

Although I think this API is better for now and to future proof, I don't feel strongly about any names or specific APIs. My goal was to try to match alloy as close as possible, while keeping necessary patterns or naming that existed before (also to minimize breaking changes like re-using `ViewCall`. Opinions or comments welcome!

This _should_ only give deprecation warnings from 0.10

Details that might be missed:
- Generic bounds loosened at type level (where it made sense to)
- `new` and `preflight` are the constructors for `Contract`. They both have the `call` method at the end of the chain, but the guest one implicitly unwraps the error. I don't love this pattern, but I'm matching the existing one
  - Happy to give argument of why we maybe should have it both return a result
- Contract and view call types moved into `contract.rs`
- `SteelEnv` trait the internal implementation is based on is kept pseudo-public, to avoid exposing the trait type to the public API. Just used to be able to use `CallBuilder` in a generic way internally.
  - This doesn't lend the APIs well for people to create their custom implementations outside of our supported ones. Intentionally so that the API surface we have to maintain is smaller, at least until someone asks for this.


steel/src/contract.rs
```diff
@@ -0,0 +1,441 @@
...
+            <Self::Db as Database>::Error: Debug;
+    }
+
+    impl<'a, H> SteelEnv<'a> for &'a GuestViewCallEnv<H>
+    where
+        H: EvmHeader,
+    {
+        type Db = WrapStateDb<'a>;
+
+        fn into_evm(self) -> Evm<'a, (), Self::Db> {
```
Wollac, line 338: In the past, the EVM builder was very likely to change in newer version also its usage is error prone. So I would rather deduplicate that part, especially since the only change is the wrapper. So personally I'd get rid of `SteelEnv`, we have two implementations of `call` anyway, so maybe we can make that distinction there.

steel/src/host/mod.rs
```diff
@@ -31,6 +31,10 @@ use revm::primitives::HashMap;
 pub mod db;
 pub mod provider;
 
+/// Alias for readability, do not make public.
+#[cfg(feature = "host")]
```
Wollac, line 35: The entire `host` module is behind that flag, so I don't think this is actually necessary.

steel/src/host/mod.rs
```diff
@@ -31,6 +31,10 @@ use revm::primitives::HashMap;
...
+pub(crate) type HostViewCallEnv<P, H> = ViewCallEnv<ProofDb<P>, H>;
```
Wollac, line 36: I like these aliases 👍

steel/src/host/db.rs
```diff
@@ -33,7 +33,7 @@ pub enum ProviderDbError<E: std::error::Error> {
 }
 
 /// A revm [Database] backed by a [Provider].
-pub struct ProviderDb<P: Provider> {
+pub struct ProviderDb<P> {
```
Wollac, line 36: You are absolutely right, technically these bounds are unnecessary and I am not sure what the most idiomatic way is. However, the `ProviderDb` has no other implementation and must always be used with a provider (which is also in the doc), so I thought it might make sense to keep them for clarity.

steel/tests/view_call.rs
```diff
@@ -249,14 +267,39 @@ fn call_eoa() {
...
+        .call_builder(&ViewCallTest::testBlockhashCall {})
+        .call()
+        .expect_err("calling an EOA should fail");
 }
 
-fn eth_call<C>(view_call: ViewCall<C>, block: u64) -> C::Return
+/// Type just used in tests to avoid restriction of operating over builders with different generic
+/// `E` types.
+#[derive(Debug, Default)]
+struct BuilderOverrides {
```
Wollac, line 280: Hmm, I think it might be even more complicated than that:
We also need the ability to specify the `with_chain_spec'. I think this is actually a shortcoming in the current tests that we do not have.

steel/src/contract.rs
```diff
@@ -0,0 +1,441 @@
...
+    }
+
+    /// Initialize a call builder to execute a call on the contract. For more information on usage,
+    /// see [Contract].
+    pub fn call_builder<C: SolCall>(&self, call: &C) -> ViewCallBuilder<&GuestViewCallEnv<H>, C> {
+        ViewCallBuilder::new_sol(self.env, self.address, call)
+    }
+}
+
+#[cfg(feature = "host")]
```
Wollac, line 83: So far I have tried to have everything that depends on the host feature in the `host` module, this also eliminates the need for those ugly configs for imports.
But of course, this approach would reduce code locality, and that is a tradeoff. So let's see how everything looks after the consolidations and see which one makes more sense.

steel/src/contract.rs
```diff
@@ -0,0 +1,441 @@
...
+        call: &C,
+    ) -> ViewCallBuilder<&mut HostViewCallEnv<P, H>, C> {
+        ViewCallBuilder::new_sol(&mut self.env, self.address, call)
+    }
+}
+
+/// A builder for calling an Ethereum contract. Once configured, call with [ViewCallBuilder::call].
+#[derive(Debug, Clone)]
+#[must_use]
+pub struct ViewCallBuilder<E, C> {
```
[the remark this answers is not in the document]
austinabell: While I agree for the longer term, I was hoping this could just be here for one version (to allow a soft deprecation) and then remove it in the following version, and I'll create the issue to track. Point being I know it's a pre-1.0, but hard breaking changes give devs a lot of overhead, so having a deprecation for a version gives them time to switch, and makes it easier.

steel/src/contract.rs
austinabell: yeah fair, I was hoping these would be more easily generic, but that does seem like a better path. I'll try that again, and either remember why I didn't go that way or I'll make that change.

steel/src/host/mod.rs
austinabell: you're right, that was a sloppy refactor :D

steel/src/host/db.rs
austinabell: Yeah, it doesn't right now, but even though that's the case, it causes very parasitic requirements of any types that use this to also define those boundaries, and gets messy for consumers in some cases. This just gives more flexibility, and future proofs imo. Happy to change back if you feel strongly about it, though.

steel/tests/view_call.rs
austinabell: Yeah, that's definitely a negative of these new APIs, that testing host and guest in the same place is messy. Although for this, the chain spec is defined per test, so this isn't a regression, just makes things a bit more verbose/messy. Is this a dealbreaker?

steel/src/contract.rs
austinabell: Agree, I'm right on the fence with this, and just made a judgement call of what _seemed_ most clean. No opinions here, so happy to switch if anyone shares an opinion!

steel/src/contract.rs
austinabell: Good callout! So originally it was needed because the call function was the same, but after that change was made, I forgot to loop back to remove :) done in [`c7f52fc` (#115)](https://github.com/risc0/risc0-ethereum/pull/115/commits/c7f52fc9374fc6414d17a10a0bd31ed2c21a0779)

steel/src/contract.rs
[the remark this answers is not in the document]
Wollac: I just talked about this in the standup.
Since this is a pretty big cleanup, we want to keep the code clean, even if we have to make these breaking changes.
We'll make sure we have examples and documentation to help users make the transition.

steel/src/host/db.rs
Wollac: No strong feeling at all. So let's keep this change.

steel/src/contract.rs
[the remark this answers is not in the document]
austinabell: done. It's much cleaner to still keep the type internally imo, so that's what I did, just removed it from public API. 

Also don't love that non-zero return check (seems like there could theoretically be a use to have a non-return function call), and I could have it in a constructor, but seemed cleanest to just have in the transact function. Noting because the error would show up when the tx is called rather than the builder initiated (I think), so wanted to point that out.

[`7a5523d` (#115)](https://github.com/risc0/risc0-ethereum/pull/115/commits/7a5523dab3177a46a882ff09d9bd2152d5dca697)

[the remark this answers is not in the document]
Wollac: > Also don't love that non-zero return check (seems like there could theoretically be a use to have a non-return function call)

There could be such a use case, but the problem is also that in `transact` we check that the function has returned, which is actually very helpful in notifying the user if they have accidentally called an EOA (a very common error), and we decode the result, which also fails for functions that do not return. Both of these things would need to be addressed before the check could be removed. Since this seems to be a rather esoteric use case, it definitely feels better to give a compile-time error than a cryptic runtime error.

steel/README.md
```diff
@@ -62,12 +65,15 @@ Here is a snippet to the [relevant code](../examples/erc20/host/src/main.rs) on
...
+    EthViewCallEnv::from_rpc(&args.rpc_url, None)?.with_chain_spec(&ETH_SEPOLIA_CHAIN_SPEC);
+let number = env.header().number();
+let commitment = env.block_commitment();
 
 // Preflight the view call to construct the input that is required to execute the function in
 // the guest. It also returns the result of the call.
-let returns = env.preflight(ViewCall::new(CALL, CONTRACT).from(CALLER))?;
+let mut contract = Contract::preflight(CONTRACT, &mut env);
+let returns = contract.call_builder(&CALL).from(CALLER).call()?;
 let input = env.into_zkvm_input()?;
```
Wollac, line 77: I think we should rename that. It seems a bit arbitrary to have `zkvm` in the name especially since there might also be other usecases of using steel. I think `into_input` is just fine.

austinabell: sure! That seems reasonable

austinabell: Didn't add a deprecation for the old naming since that was never released, but happy to if we'd like it (light weight)

steel/tests/view_call.rs
```diff
@@ -17,19 +17,34 @@
...
-        provider::{CachedProvider, EthFileProvider, EthersProvider},
-        EthersClient,
-    },
-    ViewCall,
+    host, CallBuilder, Contract,
 };
 use std::fmt::Debug;
 use test_log::test;
 
+macro_rules! provider {
```
Wollac, line 27: This was the only way I could think of to make the golden cache file generation a bit more streamlined. It also fixes the issue where the spec was not always provided (which was already an issue before this PR).
Happy to change it if there are any better suggestions....

steel/tests/view_call.rs
Wollac: I'm afraid there is no other way if we want to follow alloy's `call_builder` pattern. Still, it was much nicer before.

Wollac (approved):
We should also change the name of `EthViewCallEnv` to `EvmEnv` and `EthViewCallInput` to `EvmInput, but we can do that in another PR.

austinabell:
should I cherry-pick commits to fix the unrelated commits on this PR or is it fine the way it is? (fine by me since squashed anyway)

Wollac:
Let's just squash and merge

PR #118, opened 2024-05-16
Use recent trusted setup ceremony for stark2snark

tzerrell:
See also https://github.com/risc0/risc0/pull/1844 for the related changes in the core repo.


tzerrell:
I'm unclear about the CI error here; I'm not sure even whether it's caused by these changes. I'll dig further tomorrow.

capossele:
> I'm unclear about the CI error here; I'm not sure even whether it's caused by these changes. I'll dig further tomorrow.

No worries, the examples are broken only because they use bonsai. We switched to using a local prover in #110

nategraf:
Let's go ahead and land this. Goal is to keep `main` here up to date with `main` on `risc0`.

PR #127, opened 2024-05-30
Rename ReceiptClaimLib.from to ReceiptClaimLib.ok

nategraf:
Rename `ReceiptClaimLib.from` to `ReceiptClaimLib.ok` to match recent API changes to `risc0`.

https://github.com/risc0/risc0/blob/c393f590b68cfd1dd0d8d167b9d6833ef832d428/risc0/zkvm/src/receipt_claim.rs#L68-L73


nategraf:
@capossele, could you take a look at this? This is something I'd like to get in before 1.0, or simply drop it. Just a smaller API change to match my more recent changes to the `risc0` API.

capossele (approved):

PR #133, opened 2024-06-03
bootstrap for 1.0-rc.7 changes

capossele:
- Run bootstrap from risc0
- Manually fix testSelectorIsStable constant (should be made part of the previous bootstrap in the future)
- Add tests to check stability of `CONTROL_ROOT` and `BN254_CONTROL_ID` , so that we directly exercise those values rather than only in the `erc20-counter` example, where its failure is cryptic.


contracts/src/groth16.rs
```diff
@@ -32,3 +32,44 @@ pub fn encode(seal: Vec<u8>) -> Result<Vec<u8>> {
...
+    fn parse_digest(file_path: &str, name: &str) -> Result<String, anyhow::Error> {
+        let content = fs::read_to_string(file_path)?;
+        let re_digest = Regex::new(&format!(r#"{}\s*=\s*hex"([0-9a-fA-F]+)""#, name))?;
+        re_digest
+            .captures(&content)
+            .and_then(|caps| caps.get(1).map(|m| m.as_str().to_string()))
+            .ok_or(anyhow!("{name} not found"))
+    }
+    #[test]
+    fn control_root_is_stable() {
```
nategraf, line 58:
```suggestion
    fn control_root_is_consistent() {
```
This does seems like a good test to add. It's more about consistency than stability.

contracts/src/groth16.rs
```diff
@@ -32,3 +32,44 @@ pub fn encode(seal: Vec<u8>) -> Result<Vec<u8>> {
...
+        let params = Groth16ReceiptVerifierParameters::default();
+        let expected_control_root = params.control_root.to_string();
+        let control_root = parse_digest(CONTROL_ID_PATH, CONTROL_ROOT).unwrap();
+        assert_eq!(control_root, expected_control_root);
+    }
+
+    #[test]
+    fn bn254_control_id_is_stable() {
```
nategraf, line 66:
```suggestion
    fn bn254_control_id_is_consistent() {
```

nategraf (approved):
Looks good! Thanks for adding those tests. Definitely a good thing to check.

PR #138, opened 2024-06-04
Final updates and update version to 1.0.0

nategraf:
This PR applies final updates to the release branch for the 1.0.0 release.
* Cherry pick #129
* Cherry pick #131
* Remove or soften warnings, as is also done by https://github.com/risc0/risc0-ethereum/pull/139 on `main`

This PR bumps the version of all crates in this repo, except `risc0-steel`, to 1.0.0.


nategraf:
Bonsai staging is currently failing because it has deployed `1.0.0-rc.7` and rejects a `1.0.0` client
https://github.com/risc0/risc0-ethereum/actions/runs/9375718600/job/25814334497?pr=138

capossele:
Should we also bump the CI workflow reference of rc.7 and use 1.0.0? Pointing this out because also the examples are failing, aside from obviously Bonsai as you mentioned already

nategraf:
> Should we also bump the CI workflow reference of rc.7 and use 1.0.0? Pointing this out because also the examples are failing, aside from obviously Bonsai as you mentioned already

Ah! I missed that before. I've now updated it to use the `release-1.0` branch.

capossele (approved):

PR #140, opened 2024-06-05
Add deployment + management scripts

intoverflow:
Adds management scripts for the smart contracts.

[Quick link to readme](https://github.com/risc0/risc0-ethereum/tree/tcarstens/deploy/contracts/script)

Deployed to Sepolia:

* [Timelock](https://sepolia.etherscan.io/address/0xB4E3306129208cC8e6E75157f75f62eAe0B920a0#code)
* [Router](https://sepolia.etherscan.io/address/0x925d8331ddc0a1F0d96E68CF073DFE1d92b69187#code)
* [Estop](https://sepolia.etherscan.io/address/0xeA41bc40DAA1d0E681c7e8eA449b89a290e922e8#code)
* [Verifier](https://sepolia.etherscan.io/address/0x3d24C84FC1A2B26f9229e58ddDf11A8dfba802d0#code)

PR #145, opened 2024-06-07
Suggestions for #140

nategraf:
- rename DeployEstopVerifier and change vm.startBroadcast() and vm.broadcast()
- move common functions to base contract
- simulate scheduled calls to avoid finding out they'll fail later
- add a codeblock language
- add --broadcast flag
- fix simulation, error on unset env var

### DIFF
--- a/contracts/script/Manage.s.sol
+++ b/contracts/script/Manage.s.sol
@@ -169,9 +169,15 @@ contract ScheduleDeployEstopVerifier is RiscZeroManagementScript {
         console2.log("scheduleDelay:", scheduleDelay);
 
         bytes memory data = abi.encodeCall(verifierRouter().addVerifier, (selector, verifierEstop()));
+        address dest = address(verifierRouter());
+
+        // Simulate the call and check that it will not revert.
+        vm.prank(address(timelockController()));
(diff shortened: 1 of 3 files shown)


nategraf:
@intoverflow feel free to press the merge button if these suggestions look good to you.

contracts/script/Manage.s.sol
```diff
@@ -42,6 +42,69 @@ function timelockControllerRole(TimelockController timelockController, string me
...
+    }
+
+    /// @notice Simulates a call to check if it will succeed, given the current EVM state.
+    function simulate(address dest, bytes memory data) internal {
+        uint256 snapshot = vm.snapshot();
+        vm.prank(address(timelockController()));
+        (bool success,) = dest.call(data);
+        require(success, "simulation of transaction to schedule failed");
+        vm.revertTo(snapshot);
+    }
```
intoverflow, line 105: Brilliant

intoverflow:
This is terrific. I've tested the entire flow against Anvil and it works as advertised. I've also confirmed that every scheduled update gets simulated (using vm.prank()) before being scheduled, and that the `--broadcast` flag works as intended. Very nicely done

Issue #106, opened 2024-05-10
`sccache` not being properly invalidated on changes to `.sol` files


nategraf:
Changes to `.sol` files are not properly invalidating the `sccache`. This is likely fixed by adding a `println!` to `build.rs` to ensure that we rebuild if any of the `.sol` files that are used change.

Example of a failing run:
https://github.com/risc0/risc0-ethereum/actions/runs/9037664809/job/24837174330

PR #150, opened 2024-06-12
Add a rerun if output to try to invalidate the build cache when sol files change

nategraf:
An attempt to resolve the issues with https://github.com/risc0/risc0-ethereum/issues/106 by adding a `rerun-if` output to the `build.rs` script


nategraf:
Converted this to draft as I didn't actually get a chance to test this yet

PR #152, opened 2024-06-13
Increase compatibility with older solidity versions

capossele:
Trying to integrate and use these contracts with a project using an older Solidity version I have encountered two issues:

1. SYSTEM_STATE_ZERO_DIGEST needs to be known at compilation-time
2. The mock verifier requires a very recent solidity version


nategraf:
@capossele could you confirm when overflow checks were added?

capossele:
Looks like from 0.8.0: https://soliditylang.org/blog/2020/12/16/solidity-v0.8.0-release-announcement/

nategraf (approved):

PR #158, opened 2024-06-21
Fix publishing of `risc0-steel` (#151)

### LINKED ISSUE #144: Fix publishing of `risc0-steel`

nategraf:
This fixes #144 by copying the Solidity struct from `contracts` into the
`sol!` macro.
This removes all external dependencies, but we need to make sure that
these two places always match.

Cherry-pick of #151 to release-1.0


nategraf:
Once this is merged, I'll release a tag a new version of steel and publish to crates.io

Wollac (approved):

PR #166, opened 2024-06-27
chore: Decode MPT nodes using alloy

Wollac:
This PR moves the decoding of MPT nodes from [rlp](https://crates.io/crates/rlp) to [alloy-rlp](https://crates.io/crates/alloy-rlp).

This removes the `rlp` dependency and should be more performant. The behavior or the API of Steel is not changed.


Wollac:
Some of these changes will be integrated into `alloy-rlp` (https://github.com/alloy-rs/rlp/issues/17) lets wait for that...

capossele (approved):

PR #167, opened 2024-06-27
feat: Add hash chain support for older commitments

Wollac:
Previously, `EvmEnv::block_commitment()` returned the commitment with respect to the block holding the EVM state. This can cause problems during on-chain verification if that block is older than 256 blocks.
To address this issue, this PR adds functionality to use `EvmEnv::into_input_with_chain(self, block_number)` to prepare a hash chain from the state block to the more recent block with `block_number`. This makes it possible to generate Steel proofs for blocks older than 256 blocks.

```rust
// Create a view call environment from an RPC endpoint and a block number in the past.
let mut env = EthViewCallEnv::from_rpc(&args.rpc_url, Some(old_block_number))?;

// Preflight the call to construct the input with respect to a more recent block.
let mut contract = Contract::preflight(CONTRACT, &mut env);
let returns = contract.call_builder(&CALL).from(CALLER).call()?;
let input = env.into_input_with_chain(latest_block_number)?;
```

### DIFF
--- a/steel/src/host/mod.rs
+++ b/steel/src/host/mod.rs
@@ -19,7 +19,7 @@ use self::{
     provider::{EthersProvider, Provider},
 };
 use crate::{ethereum::EthEvmEnv, EvmBlockHeader, EvmEnv, EvmInput, MerkleTrie};
-use alloy_primitives::{Sealable, B256};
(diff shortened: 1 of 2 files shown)


Wollac:
Superseded by #170

Issue #174, opened 2024-07-11
[BUG] groth16::encode takes a owned Vec<u8> when it could take a slice


mothran:
## Bug Report

https://github.com/risc0/risc0-ethereum/blob/main/contracts/src/groth16.rs#L25

In the groth16::encode() method it would be nice if the seal param was a slice instead of a Vec so I could avoid a clone pulling the seal from a receipt.

nategraf:
Closed by https://github.com/risc0/risc0-ethereum/pull/179

PR #175, opened 2024-07-15
Use `impl AsRef<u8>` instead of `Vec` for encode Groth16 args

nategraf:
Marked as draft as I haven't had a chance to check whether this would be a breaking change yet.

Resolves: #174


capossele:
Closing as integrated into #179

PR #179, opened 2024-07-18
Improve groth16 encode API

### LINKED ISSUE #174: [BUG] groth16::encode takes a owned Vec<u8> when it could take a slice

### LINKED ISSUE #176: Unintuitive `groth16::encode` API
Currently, it's unintuitive that from a receipt you need to know about this method, as well as the fact that you have to use a git dependency of `risc0-ethereum` since not yet released.

Currently the usage looks like this:
https://github.com/risc0/risc0-foundry-template/blob/f8671069cd5f7ea54743e1007f2c26acfd0c80ad/apps/src/bin/publisher.rs#L130

and the API is _slightly_ improved by removing the need for a clone in #175 / #174 but the API is still based around an ambiguous `[u8]` which is unclear what that represents, as well as the return value also being `Vec<u8>`.

The functions also return a `Result` which I assume is for future proofing it, but confusing to me personally because there doesn't seem to be anything than can error in the future.

Possibly an opinionated point to have the param be more strictly typed or have the API based around the receipt type, but it seems the less opinionated action items might be:
- Have this method be released in some version, either this crate separately, or looped into the zkvm crate under some feature flag (to avoid adding deps).
  - Latter might not be possible with circular deps, but quick scan does seem like it can be
- Document where the seal can be pulled from and examples in docs
  - https://github.com/risc0/risc0-ethereum/blob/bfea3b870fe22f05ddd81093d84a76ce1abf9eb1/contracts/src/groth16.rs#L24 I personally think would be a confusing explanation to someone unfamiliar?

capossele:
Pending #185 

Attempt at solving #174 and #176. 

Closes #174
Closes #176


contracts/src/groth16.rs
```diff
@@ -12,35 +12,34 @@
...
 }
 
-/// encoding of the seal with selector.
-pub fn encode(seal: Vec<u8>) -> Result<Vec<u8>> {
-    let verifier_parameters_digest = Groth16ReceiptVerifierParameters::default().digest();
-    let selector = &verifier_parameters_digest.as_bytes()[..4];
+/// Encoding of a [Groth16Receipt][risc0_zkvm::Groth16Receipt] by prefixing it with the verifier selector,
+/// taken from the first 4 bytes of the hash of the verifier parameters including the Groth16
+/// verification key and the control IDs that commit to the RISC Zero circuits.
+pub fn encode(receipt: &Groth16Receipt<ReceiptClaim>) -> Vec<u8> {
```
austinabell, line 26: How will we recommend going from https://docs.rs/bonsai-sdk/0.9.0/bonsai_sdk/responses/struct.SnarkReceipt.html, given we still recommend using bonsai-sdk over the zkvm for production use cases?

nategraf (approved):
The new method for encoding a seal on input of a `Receipt` is the right direction. I don't think we should introduce a new type though. I've made some modifications and rebased this PR onto https://github.com/risc0/risc0-ethereum/pull/183

PR #180, opened 2024-07-19
WEB3-68: feat: Add EIP4788 commitments

Wollac:
Adds the option to commit to a Beacon (Consensus) block instead of an Execution block.


nategraf:
Marked draft, as it is dependent on switching from Ethers to Alloy. Once that PR is in, this PR will be ready to go.

examples/beacon-root/contracts/src/Beacon.sol
```diff
@@ -0,0 +1,54 @@
...
+    /// @dev https://eips.ethereum.org/EIPS/eip-4788
+    address internal constant BEACON_ROOTS_ADDRESS = 0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02;
+
+    /// @notice The Beacon block root could not be found as the next block has not been issued yet.
+    error NoParentBeaconBlock();
+
+    /// @notice Attempts to find the root of the Beacon block with the given timestamp.
+    /// @dev Since the Beacon roots contract only returns the parent Beacon block’s root, we need to find the next
+    ///      Beacon block instead. This is done by adding the block time of 12s until a value is returned.
+    function blockRoot(uint256 timestamp) internal view returns (bytes32 root) {
```
nategraf, line 33: When would looking up by timestamp be used instead of by hash?

nategraf (commented):
As we discussed on Slack, it would be nice to integrate this such that it's easy to switch from execution block commitments to consensus. In particular, it would be nice to avoid maintaining a separate example. Do you think this is possible in this PR?

examples/beacon-root/contracts/src/Beacon.sol
Wollac: I am not sure I understand. The EIP-4788 beacon root contract gives you a way to retrieve the (parent) beacon block root by its respective timestamp. As such, Steel commits to the beacon block root plus the beacon block timestamp, and during on-chain validation we check that the beacon root contract returns the same root for that timestamp.

examples/beacon-root/contracts/src/Beacon.sol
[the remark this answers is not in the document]
Wollac: Yes 100% it should. I just added it to the example for now to run some tests with it and figure out the final interface before moving it.

Cargo.toml
```diff
@@ -32,6 +32,10 @@ alloy-sol-types = { version = "0.7" }
 alloy = { version = "0.2.1", features = ["full"] }
 alloy-trie = { version = "0.4.0" }
 
+# Beacon chain support
```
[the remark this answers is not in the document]
Wollac: See my comments on that issue.

examples/beacon-root/contracts/src/Beacon.sol
nategraf: Ok, thank you. That clarifies my question.

examples/erc20-counter/contracts/test/Counter.t.sol
```diff
@@ -0,0 +1,93 @@
...
+import {Steel, Beacon, Encoding} from "risc0/steel/Steel.sol";
+import {ERC20} from "openzeppelin-contracts/token/ERC20/ERC20.sol";
+
+contract ERC20FixedSupply is ERC20 {
+    constructor(string memory name, string memory symbol, address owner) ERC20(name, symbol) {
+        _mint(owner, 1000);
+    }
+}
+
+contract CounterTest is Test {
```
Wollac, line 33: These tests are actually never run in the CI. Should we change that or at least move the tests of the commitment validation into the general risc0-ethereum foundry tests

examples/README.md
```diff
@@ -11,12 +11,17 @@ Explore a more advanced interaction between [Steel] and a custom Ethereum smart
 - Verify ERC20 token ownership (minimum 1 token required) before incrementing.
 - Leverage RISC Zero as a [coprocessor] for efficient proof generation and verification.
 
+## [ERC20 Counter using EIP-4788 Commitment Validation](./eip4788-counter/README.md)
+
+This example is very similar to the previous ERC20 Counter Example. The only difference being that [Steel] is used to produce a commitment with respect to the Beacon block root instead of the Execution block hash.
+This commitment is using the [EIP-4788] Beacon roots contract.
+
```
nategraf, line 18: It looks like this needs to be updated.

examples/erc20-counter/apps/src/bin/publisher.rs
```diff
@@ -16,61 +16,69 @@
...
-    /// Ethereum Node endpoint.
+    /// Ethereum RPC endpoint URL
     #[clap(long, env)]
-    rpc_url: Url,
+    eth_rpc_url: Url,
 
-    /// Counter's contract address on Ethereum
-    #[clap(long)]
-    contract: Address,
+    /// Beacon API endpoint URL
```
nategraf, line 69:
```suggestion
    /// Optional Beacon API endpoint URL.
    ///
    /// If provided, the Steel proof will use a beacon block commitment, rather than the execution block.
    /// This allows proofs to be validated against a block hash from EIP-4788 with n expiration of about 1 day,
    /// rather than an execution block hash with an expiration of about 50 minutes.
```

examples/erc20-counter/contracts/test/Counter.t.sol
nategraf: I'd like to have these tests running in CI if at all possible. Does this run locally in Anvil alright, considering the use of `Beacon.blockRoot`?

examples/erc20-counter/contracts/test/Counter.t.sol
```diff
@@ -0,0 +1,93 @@
...
+        // create a mock proof
+        RiscZeroReceipt memory receipt = verifier.mockProve(imageId, sha256(abi.encode(journal)));
+
+        uint256 previous_count = counter.get();
+
+        counter.increment(abi.encode(journal), receipt.seal);
+
+        // check that the counter was incremented
+        assert(counter.get() == previous_count + 1);
+    }
```
nategraf, line 92: This is actually quite a nice example of how to use the mock verifier in a test. A good example for others to use 👌

steel/src/lib.rs
```diff
@@ -179,3 +128,72 @@ pub trait EvmBlockHeader: Sealable {
...
+            /// Encodes both the block identifier (block number or timestamp) and the version.
+            uint256 blockID;
+            /// The block hash or beacon block root, used for validation.
+            bytes32 blockDigest;
+        }
+    }
+}
+
+/// Solidity struct representing the committed block used for validation.
+pub use private::Commitment as SolCommitment;
```
nategraf, line 147: Since there is not another type named `Commitment`, I would lean towards making this publicly available with that name rather than `SolCommitment`. Based on this name, I figured this was the "Solidity version" of some other Rust type and was surprised not to find that other type anywhere.

nategraf (approved):
Overall, looks good to me!

examples/erc20-counter/contracts/test/Counter.t.sol
Wollac: The beacon commitment test relies on anvil fork mode and thus an RPC endpoint. However, using a publicnode.com I believe would be fine, even in the CI.

steel/src/lib.rs
Wollac: Yeah, I see what you mean. I thought it might make sense to use this Sol* name to indicate that the struct has been generated using the `sol!` macro, as for those the type and especially the naming could be strange. It is a shame that the macro does not convert snake_case to camelCase automatically.

Wollac: However, I don't have a strong opinion on this at all. I am fine with removing this renaming if we think it makes the naming cleaner. I am also not a fan of the `mod private`, as it is not really necessary.

steel/src/lib.rs
```diff
@@ -32,79 +36,24 @@ mod state;
...
-/// The serializable input to derive and validate a [EvmEnv].
+/// The serializable input to derive and validate an [EvmEnv] from.
 #[derive(Clone, Serialize, Deserialize)]
-pub struct EvmInput<H> {
-    header: H,
-    state_trie: MerkleTrie,
-    storage_tries: Vec<MerkleTrie>,
-    contracts: Vec<Bytes>,
-    ancestors: Vec<H>,
+pub enum EvmInput<H> {
```
Wollac, line 41: Should we mark this as `#[non_exhaustive]`, which should theoretically allow us to add more input types in the future as a non-breaking change.

examples/erc20-counter/contracts/test/Counter.t.sol
nategraf: I'm happy to give it a try using publicnode. Hopefully it'll be reliable and that'll just solve the issue 👌

steel/src/lib.rs
nategraf: Good idea. I went ahead and did that in https://github.com/risc0/risc0-ethereum/pull/180/commits/9fe9316085da9004ad5e11914417a22c6e4961f4

steel/src/lib.rs
nategraf: I went ahead and renamed `SolCommitment` to `Commitment`. I did not remove the `private` mod as I just wasn't sure if there are perhaps more symbols in that module that would get exported.

PR #183, opened 2024-07-24
WEB3-67: feat: switch Steel to alloy

Wollac:
### ⚡️ Features

- Replace `ethers` dependency completely with `alloy`.
- Make `host` functions `async`.
- Add support to build `EvmEnv` from any `alloy` provider.
- Add more efficient RLP-based serialization for the header.

### 🛠 Fixes

- Store the commitment inside the `EvmEnv`.
- Use `eth_getTransactionCount` and `eth_getBalance` instead of `eth_getProof` to query basic account information.
- Switch tests from pre-recorded RPC responses to `Anvil`.

### 🚨 Breaking Changes

- `EthEvmEnv::from_rpc` now accepts a `Url` instead of a `&str` for the HTTP RPC endpoint.
- `EvmEnv::from_provider` now requires an `alloy` provider, and the block number parameter has been changed to a `BlockNumberOrTag`.
- `EvmEnv::sol_commitment` has been replaced with `EvmEnv::commitment` (to get a reference), or `EvmEnv::into_commitment` (to consume and return the commitment).
- `ETH_SEPOLIA_CHAIN_SPEC` and `ETH_MAINNET_CHAIN_SPEC` have been moved to the `ethereum` module.
- `CachedProvider` has been removed completely. As alternatives, you can:
  - Use `anvil --fork-url https://ethereum-rpc.publicnode.com@20475759` to create a cached fork for block `20475759`. 
  - Cache the RPC responses on an HTTP level using [Tower](https://crates.io/crates/tower) or a caching forward proxy.
- The host functions are now `async` instead of blocking:
```rust
// Create an EVM environment from an RPC endpoint and a block number or tag.
let mut env = EthEvmEnv::from_rpc(args.rpc_url, BlockNumberOrTag::Latest).await?;
//  The `with_chain_spec` method is used to specify the chain configuration.
env = env.with_chain_spec(&ETH_SEPOLIA_CHAIN_SPEC);

// Preflight the call to prepare the input that is required to execute the function in
// the guest without RPC access. It also returns the result of the call.
let mut contract = Contract::preflight(CONTRACT, &mut env);
let returns = contract.call_builder(&CALL).from(CALLER).call().await?;

// Finally, construct the input from the environment.
let input = env.into_input().await?;
```


capossele (approved):

Wollac:
We still need to upgrade alloy to `v0.2.1`.

steel/src/host/mod.rs
```diff
@@ -13,86 +13,106 @@
...
-        let provider = EthersProvider::new(client);
-
-        // get the latest block number if none is provided
-        let block_number = match block_number {
-            Some(n) => n,
-            None => provider.get_block_number()?,
-        };
-
-        EvmEnv::from_provider(provider, block_number)
+    pub async fn from_rpc(url: Url, number: BlockNumberOrTag) -> anyhow::Result<Self> {
```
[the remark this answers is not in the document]
nategraf: It's possible. One API solution to this is to use a builder pattern and split the method that adds the RPC URL from the method that specifies the block ref. In this case, we could provide two methods, one for `BlockNumberOrTag` and one for block hash (or something like that).

steel/src/state.rs
```diff
@@ -0,0 +1,207 @@
...
+pub struct StateDb {
+    /// State MPT.
+    state_trie: MerkleTrie,
+    /// Storage MPTs to their root hash.
+    /// [Rc] is used fore MPT deduplication.
+    storage_tries: HashMap<B256, Rc<MerkleTrie>>,
+    /// Contracts by their hash.
+    contracts: HashMap<B256, Bytes>,
+    /// Block hashes by their number.
+    block_hashes: HashMap<u64, B256>,
```
nategraf, line 42: We might want to use `BTreeMap` here. I've seen it perform much better in the zkVM.

steel/src/state.rs
```diff
@@ -0,0 +1,207 @@
...
+}
+
+impl StateDb {
+    /// Creates a new state database from the given tries.
+    pub fn new(
+        state_trie: MerkleTrie,
+        storage_tries: impl IntoIterator<Item = MerkleTrie>,
+        contracts: impl IntoIterator<Item = Bytes>,
+        block_hashes: HashMap<u64, B256>,
```
nategraf, line 51: Should this also be `impl IntoIterator` or is there a reason not to (e.g. maybe it would waste a lot of time constructing the map)?

steel/tests/steel.rs
```diff
@@ -0,0 +1,236 @@
...
...01b8152600401602060405180830381865afa1580156101b8573d5f803e3d5ffd5b505050506040513d601f19601f820116820180604052508101906101dc91906102b3565b7f00000000000000000000000000000000000000000000000000000000000000006001600160a01b0316633fa4f2456040518163ffffffff1660e01b8152600401602060405180830381865afa158015610238573d5f803e3d5ffd5b505050506040513d601f19601f8201168201806040525081019061025c91906102b3565b61026691906102ca565b61027091906102ca565b905090565b6040515f906002906020818481855afa158015610294573d5f803e3d5ffd5b5050506040513d601f19601f8201168201806040525081019061027091905b5f602082840312156102c3575f80fd5b5051919050565b808201808211156102e957634e487b7160e01b5f52601160045260245ffd5b9291505056fea2646970667358221220193bf7ba8b1afe2d8e735c81e13a2034829eae79d36eac8db076b92ff6308ba564736f6c634300081900336080604052348015600e575f80fd5b5060405160c938038060c9833981016040819052602991602f565b5f556045565b5f60208284031215603e575f80fd5b5051919050565b60798060505f395ff3fe6080604052348015600e575f80fd5b50600436106026575f3560e01c80633fa4f24514602a575b5f80fd5b60315f5481565b60405190815260200160405180910390f3fea2646970667358221220c28e11a00819347c873d20f4ecc0a120e64201cb1a824335d007cf34e8c3ff8164736f6c63430008190033")]
```
nategraf, line 39: How do these values get updated?

steel/tests/erc20.rs
```diff
@@ -0,0 +1,94 @@
...
...fffffffffffffffffffffffffffffffffff168152602001908152602001600020819055506000809054906101000a900473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff163373ffffffffffffffffffffffffffffffffffffffff167fddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef856040518082815260200191505060405180910390a35b8473ffffffffffffffffffffffffffffffffffffffff163373ffffffffffffffffffffffffffffffffffffffff167fddf252ad1be2c89b69c2b068fc378daa952ba7f163c4a11628f55a4df523b3ef846040518082815260200191505060405180910390a35050505050565b6000806000841415612a185760009150612a37565b8284029050828482811515612a2957fe5b04141515612a3357fe5b8091505b5092915050565b6000600560008473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002060008373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002054905092915050565b6000808284811515612ad357fe5b0490508091505092915050565b6000828211151515612aee57fe5b818303905092915050565b6000808284019050838110151515612b0d57fe5b80915050929150505600a165627a7a72305820645ee12d73db47fd78ba77fa1f824c3c8f9184061b3b10386beb4dc9236abb280029")]
```
nategraf, line 29: How do these values get updated?

steel/tests/steel.rs
```diff
@@ -0,0 +1,236 @@
...
+        test_provider().await,
+        STEEL_TEST_CONTRACT,
+        SteelTest::testEoaAccountCall {},
+        CallOptions::new(),
+    )
+    .await;
+    assert_eq!(result.size, uint!(0_U256));
+}
+
+#[tokio::test] // Anvil crashes, if logging is enabled
```
nategraf, line 154: If this is something we can address, can you create a quick issue it and add it's number here?

nategraf (approved):
Nice work! I added some comments. None of them are concerns about the core of this PR.

steel/tests/erc20.rs
Wollac: This one is the bytecode of the contract as deployed at 0xdAC17F958D2ee523a2206206994597C13D831ec7. Fortunately, it will never change.

steel/tests/steel.rs
Wollac: Seems like this was somehow fixed in the latest version, so now even logging works here.

steel/src/config.rs
```diff
@@ -159,6 +108,18 @@ impl ChainSpec {
 mod tests {
     use super::*;
 
+    use once_cell::sync::Lazy;
+
+    static ETH_MAINNET_CHAIN_SPEC: Lazy<ChainSpec> = Lazy::new(|| ChainSpec {
```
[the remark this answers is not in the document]
Wollac: I tried to keep the tests independent of possible changes to the config. I've now taken the opportunity to clean everything up and use artificial test data instead.

examples/erc20-counter/apps/src/bin/publisher.rs
```diff
@@ -66,18 +75,24 @@ struct Args {
...
 
-    // Create an EVM environment from an RPC endpoint and a block number. If no block number is
-    // provided, the latest block is used.
-    let mut env = EthEvmEnv::from_rpc(&args.rpc_url, None)?;
+    // Create an alloy provider for that private key and URL.
+    let wallet = EthereumWallet::from(args.eth_wallet_private_key);
+    let provider = ProviderBuilder::new()
+        .with_recommended_fillers()
+        .wallet(wallet)
+        .on_http(args.rpc_url);
```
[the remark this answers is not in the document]
Wollac: Very interesting, I didn't even know of this one. It even takes a string similar to the eth provider before...

 🤔 I kinda liked the change to Url for its additional type safety and I guess HTTP is the more realistic provider as we don't rely on any pubsub and even then we have the `from_provider` for any alloy provider...

steel/src/state.rs
Wollac: I did some quick Benchmarks and at least for this very particular usecase where the `block_hashes` never contains more than 256 elements. The performance was more or less identical. The initialisation of the HashMap was a tiny bit faster while the getting was a tiny bit faster for the BTreeMap.
I guess in this case I'll leave it at the `HashMap` because of the better theoretical performance.

steel/src/state.rs
Wollac: The HashMap is actually already constructed during validation in the `into_env` step. So I think it makes sense to just move it here.

examples/erc20-counter/apps/src/bin/publisher.rs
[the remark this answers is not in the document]
nategraf: On the question of using `Url`, I also like that from a semantics point of view. With `on_builtin`, we could call `url.as_str` to pass it in, so we can keep that change either way. On the use of HTTP provider. I do see that we constrain the types to HTTP providers anyway. Something we should think about before marking 1.0 is whether we should/can loosen thew type constraints there. It's really not a major issue in any case.

steel/src/state.rs
nategraf: Ya, I don't think it'll matter either way then here 👍

steel/tests/steel.rs
nategraf: Could you add a quick note on how to do that here? I wouldn't know what to do.

steel/tests/erc20.rs
nategraf: Can you still add a note on how you got it in the first place?

Wollac: I have added the Etherscan link to that address

examples/erc20-counter/apps/src/bin/publisher.rs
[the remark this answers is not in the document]
Wollac: The types of `EthEvmEnv::from_rpc` are only `Http` because this is what `.on_http` returns. There is also https://github.com/risc0/risc0-ethereum/blob/eded136a25ff400e3f108da9ad17ae51e4c86fbd/steel/src/host/mod.rs#L62
which works with any generic alloy provider.

PR #185, opened 2024-08-05
bootstrap groth16

capossele:


nategraf (approved):
Thanks for taking care of this

Wollac (approved):
CI should be fixed with https://github.com/risc0/risc0/pull/2189

PR #187, opened 2024-08-07
WEB3-66: Add version string to RiscZeroGroth16Verifier.sol

nategraf:
In order to make it easier to determine which version a given verifier contract is, this PR adds a semantic version string.

Note that this does mean an additional step in the release process. I couldn't think of a way to make this automatic.


capossele (approved):

contracts/src/groth16/RiscZeroGroth16Verifier.sol
```diff
@@ -53,6 +53,9 @@ contract RiscZeroGroth16Verifier is IRiscZeroVerifier, Groth16Verifier {
     using OutputLib for Output;
     using SafeCast for uint256;
 
+    /// Semantic version of the the RISC Zero system of which this contract is part.
+    string constant VERSION = "1.1.0-alpha.1";
```
capossele, line 57: I wonder though, let's say no changes in the verifier contract between  x.x.0 and x.x.5 are required, does that mean the deployed contract will keep an older version than the latest crate version?

nategraf: That's a good question. I think both changing it or not changing it is fine. Mostly it's there to make it easier to find the source code in GitHub when you are looking at the contract source in Etherscan.

PR #190, opened 2024-08-09
Add dependency version check

capossele:
Add CI script to help checking for dependencies version consistencies across workspaces


capossele:
closing for now as this approach is not reliable

PR #197, opened 2024-08-18
WEB3-79: Refactor of Governance Example for RISC Zero zkVM 1.0

sashaaldrick:
This is a refactor of the original governance example located in release-0.20 [here](https://github.com/risc0/risc0/tree/release-0.20/bonsai/examples/governance). This has been rewritten to match the foundry template layout, amongst many other features.

At a high level, this PR:
- upgrades governance example guest to zkVM 1.0
- deprecates Relay workflow and contracts
- updates publisher to handle deprecated relay workflows
- simplifies test logic for readability
- creates gas benchmarks from these tests

For a high level walkthrough, please see [README.md](https://github.com/risc0/risc0-ethereum/blob/a5ccf7872c3d8dfe86d42816b2388da4e27af4c3/examples/governance/README.md), and to run tests locally with benchmarking, see [instructions.md](https://github.com/risc0/risc0-ethereum/blob/a5ccf7872c3d8dfe86d42816b2388da4e27af4c3/examples/governance/instructions.md).


sashaaldrick:
Lots of CI issues, I will take a look at a stab at fixing these later today/early this coming week

sashaaldrick:
okay I fixed some basic CI issues due to licensing and some formatting, the one that's stumping me is from the action `cargo-risczero-install`. Specifically the error is:

```sh
ompiling risc0-zkvm v1.2.0-alpha.1 (/opt/actions-runner/_work/risc0-ethereum/risc0-ethereum/tmp/risc0/risc0/zkvm)
error[E0599]: no method named `env` found for struct `Arg` in the current scope
  --> risc0/cargo-risczero/src/commands/datasheet/mod.rs:41:38
   |
41 |     #[arg(long, value_name = "PATH", env = "RISC0_SERVER_PATH")]
   |                                      ^^^ method not found in `Arg`

For more information about this error, try `rustc --explain E0599`.
error: could not compile `cargo-risczero` (lib) due to 1 previous error
error: failed to compile `cargo-risczero v1.2.0-alpha.1 (/opt/actions-runner/_work/risc0-ethereum/risc0-ethereum/tmp/risc0/risc0/cargo-risczero)`, intermediate artifacts can be found at `/opt/actions-runner/_work/risc0-ethereum/risc0-ethereum/tmp/risc0/target`.
To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
```

I thought it was something to do with clap and perhaps using a new version, I see that we are `cargo-risczero v1.2.0-alpha.1` here which perhaps has something to do with it 🤔

examples/governance/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+bytemuck = { version = "1.14" }
+ethers = { version = "2.0" }
+hex = { version = "0.4" }
+log = { version = "0.4" }
+methods = { path = "./methods" }
+risc0-build = { version = "1.0", features = ["docker"] }
+risc0-build-ethereum = { git = "https://github.com/risc0/risc0-ethereum", tag = "v1.0.0" }
+risc0-ethereum-contracts = { git = "https://github.com/risc0/risc0-ethereum", tag = "v1.0.0" }
+risc0-zkvm = { version = "1.0", default-features = false }
+risc0-zkp = { version = "1.0", default-features = false }
```
capossele, line 24:
```suggestion
risc0-build = { git = "https://github.com/risc0/risc0", branch = "main", features = ["docker"] }
risc0-build-ethereum = { path = "../../build" }
risc0-ethereum-contracts = { path = "../../contracts" }
risc0-zkvm = { git = "https://github.com/risc0/risc0", branch = "main", default-features = false }
risc0-zkp = { git = "https://github.com/risc0/risc0", branch = "main", default-features = false }
```

examples/governance/apps/src/bin/publisher.rs
```diff
@@ -0,0 +1,175 @@
...
+// limitations under the License.
+
+// This application demonstrates how to send an off-chain proof request
+// to the Bonsai proving service and publish the received proofs directly
+// to your deployed app contract.
+
+use alloy_sol_types::{sol, SolInterface};
+use anyhow::{Context, Result};
+use clap::Parser;
+use ethers::prelude::*;
```
capossele, line 22: One suggestion I have is to drop `ethers` and instead use `alloy`. We need to do the same in the foundry-template as well. For now you could do something similar as in https://github.com/risc0/risc0-ethereum/blob/main/examples/erc20-counter/apps/src/bin/publisher.rs

examples/governance/apps/src/bin/publisher.rs
```diff
@@ -0,0 +1,175 @@
...
+    // Compute the ImageID of `finalize_votes.rs` ELF binary.
+    // let image_id = compute_image_id(FINALIZE_VOTES_ELF)?;
+    // let image_id_bytes: [u8; 32] = image_id
+    //     .as_bytes()
+    //     .try_into()
+    //     .expect("Digest should be 32 bytes");
+    // let image_id_fixed_bytes = FixedBytes::<32>::from(image_id_bytes);
+
+    // Encode the seal with the selector.
+    let seal = groth16::encode(receipt.inner.groth16()?.seal.clone())?;
```
capossele, line 153:
```suggestion
    let seal = risc0_ethereum_contracts::encode_seal(&receipt)?;
```

Using this new method allows you to be compatible with `RISC0_DEV_MODE` as well

sashaaldrick: that's much better and makes sense thanks @capossele

examples/governance/apps/src/bin/publisher.rs
sashaaldrick: thank you for noticing that, I will definitely do that

sashaaldrick:
I'm working on refactoring the `publisher.rs` to work like the publisher in the erc20 counter, i.e using alloy directly instead of the `TxSender` implementation.

examples/governance/apps/src/bin/publisher.rs
```diff
@@ -0,0 +1,175 @@
...
+    let tx_sender = TxSender::new(
+        args.chain_id,
+        &args.rpc_url,
+        &args.eth_wallet_private_key,
+        &args.contract,
+    )?;
+
+    // Decode the hex-encoded proposal ID and votes data
+    let proposal_id = decode(&args.proposal_id).context("Failed to decode proposal ID")?;
+    let votes_data = decode(&args.votes_data).context("Failed to decode votes data")?;
```
nategraf, line 119: In cases like this, I like to use the fully qualified name. It saves the effort of having to look for which `decode` method this is.
```suggestion
    let proposal_id = hex::decode(&args.proposal_id).context("Failed to decode proposal ID")?;
    let votes_data = hex::decode(&args.votes_data).context("Failed to decode votes data")?;
```

examples/governance/apps/src/bin/publisher.rs
```diff
@@ -0,0 +1,175 @@
...
+        )?
+        .receipt;
+
+    // Compute the ImageID of `finalize_votes.rs` ELF binary.
+    // let image_id = compute_image_id(FINALIZE_VOTES_ELF)?;
+    // let image_id_bytes: [u8; 32] = image_id
+    //     .as_bytes()
+    //     .try_into()
+    //     .expect("Digest should be 32 bytes");
+    // let image_id_fixed_bytes = FixedBytes::<32>::from(image_id_bytes);
```
nategraf, line 150: Why would you do this instead of using `methods::FINALIZE_VOTES_ID`?

examples/governance/apps/src/bin/publisher.rs
```diff
@@ -0,0 +1,175 @@
...
+    #[clap(long)]
+    contract: String,
+
+    /// The proposal ID (32 bytes, hex-encoded)
+    #[clap(long)]
+    proposal_id: String,
+
+    /// The votes data (hex-encoded, multiple of 100 bytes)
+    #[clap(long)]
+    votes_data: String,
```
nategraf, line 101: How are folks expected to construct the vote data? In an end-to-end application, this would be something that is constructed by pulling event data. It would be useful to show that pattern if its not too much to implement.

examples/governance/methods/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+This directory contains the [zkVM] portion for the RISC Zero Governance application. 
+To learn more about the code in this directory, we recommend [Guest Code 101] and the [Foundry Template]. 
+
+[Guest Code 101]: https://dev.risczero.com/api/zkvm/guest-code-101
+[zkVM]: https://dev.risczero.com/zkvm
+[Foundry Template]: https://github.com/risc0/risc0-foundry-template
+
+
+
```
nategraf, line 11:
```suggestion
```

nategraf (commented):
Here is a partial review. I need to come back and get through the rest.

examples/governance/apps/src/bin/publisher.rs
sashaaldrick: I assume you mean the commented out code, that is a remnant from the old example _I think_. This will be removed before the next review, thanks

examples/governance/apps/src/bin/publisher.rs
sashaaldrick: I have added a `generate_proposal_id` and `generate_vote_data`, now there are a _lot_ of caveats, as the signed digest includes a nonce, address of the deployed contract etc. but it's a start for people

examples/governance/contracts/utils/ImageID.sol
nategraf, line 1: We don't include this file since it's automatically generated by `cargo build`. It would be nice to have it be possible to clone this repo and run `forge test`. However, this is not possible because the guest ELF will not be available until `cargo build` is run. Is there another reason to check this file in to `git`? If not, I'd recommend continuing to exclude it.

examples/governance/methods/guest/Cargo.toml
nategraf, line 1: We should use ECDSA acceleration here, unless there is some reason not to. Here is an example of how to enable it. `0.13.3` is the latest version we have patch support for at the moment.

I'm also surprised the `[workspace]` line is excluded. Did you do that on purpose?

https://github.com/risc0/risc0/blob/victor/local-proving-instructions/examples/ecdsa/methods/guest/Cargo.toml

examples/governance/methods/guest/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+  "std",
+  "ecdsa",
+], default-features = false }
+risc0-zkvm = { workspace = true, default-features = false, features = ["std"] }
+tiny-keccak = { version = "2.0", features = ["keccak"] }
+
+[profile.release]
+lto = "thin"
+opt-level = 3
+debug = true
```
nategraf, line 24: `opt-level = 3` is not required, as it is the default. Adding `codegen-units = 1` is generally helpful.
```suggestion
# Empirically observed to result in the best performance for this binary (in particular, better than
# "fat"). It's often a good idea to experiment with different optimization levels and options.
codegen-units = 1
lto = "thin"
debug = true
```

examples/governance/script/GenerateVoteData.s.sol
nategraf, line 1: This file is doesn't make sense. It seems to mostly be copy-pasted from `Deploy.s.sol`, but the it's unclear if it's supposed to be a deployment script. It appears to instead be a script to gather the data needed to construct the zkVM input. Is that accurate?

examples/governance/tests/BaselineGovernorTest.sol
```diff
@@ -0,0 +1,258 @@
...
+
+        assertGt(proposalId, 0, "Proposal should be created with non-zero ID");
+        assertEq(
+            uint256(baselineGovernor.state(proposalId)),
+            uint256(IGovernor.ProposalState.Pending),
+            "New proposal should be in Pending state"
+        );
+    }
+
+    function testProposalThreshold() public {
```
nategraf, line 53: Is the name of this test correct? It looks like this test is actually checking that different proposals have different IDs.

examples/governance/tests/RiscZeroGovernorTest.sol
```diff
@@ -0,0 +1,332 @@
...
+        assertEq(
+            uint256(riscZeroGovernor.state(proposalId)),
+            uint256(IGovernor.ProposalState.Executed),
+            "Proposal should be executed"
+        );
+
+        vm.clearMockedCalls();
+    }
+
+    function testFailToReachQuorum() public {
```
nategraf, line 304: Because `riscZeroGovernor.verifyAndFinalizeVotes` is not called, this test would fail even if enough votes where cast.

examples/governance/tests/benchmarks/BenchmarkTestBase.sol
```diff
@@ -0,0 +1,169 @@
...
+    ) internal noGasMetering returns (bytes32 journalDigest) {
+        bytes memory journal = getJournal(proposalId, noOfAccounts);
+        journalDigest = sha256(journal);
+    }
+
+    function hashBallots(
+        uint256 noOfAccounts
+    ) internal noGasMetering returns (bytes32, bytes memory) {
+        bytes memory encodedBallots;
+        bytes32 ballotHash = 0x296dc540e823507aa12a2e7be3c9c01672a7d9bb7840214223e8758fdb2986c7;
```
nategraf, line 129: Where does this value come from? Could you add a comment or generate it automatically?

nategraf (commented):
Looking pretty good. Some things don't quite make sense to me.

I've also opened a PR with some minor suggestions. Feel free to merge it if it looks good to you. https://github.com/risc0/risc0-ethereum/pull/226

examples/governance/apps/README.md
```diff
@@ -0,0 +1,88 @@
+# Apps
+
+In typical applications, an off-chain app is needed to do two main actions:
+
+* Produce a proof (see [proving options]).
```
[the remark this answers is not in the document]
sashaaldrick: Ah now I understand, thanks for that. I know Nuke had some discussions with Frank about [lychee](https://github.com/lycheeverse/lychee) and that's now used in the main risc0 monorepo CI iirc, perhaps for consistency, it would make sense to use it in this repo as well

examples/governance/contracts/utils/ImageID.sol
sashaaldrick: I think I had originally removed it, and for some testing reasons I had let it back in. I will remove it again, thanks.

examples/governance/script/GenerateVoteData.s.sol
sashaaldrick: Yes this was a test I had run, thanks for catching that

examples/governance/tests/benchmarks/BenchmarkTestBase.sol
sashaaldrick: alright, I took a look and it was actually very easy to find that there is a salt

```solidity
        // Write the proposal id to the ballot box commitment to act as a salt.
        _proposalVotes[id].ballotBoxCommit = bytes32(id);
```

I will generate this automatically now, thanks Victor!

examples/governance/tests/RiscZeroGovernorTest.sol
sashaaldrick: ah, I thought I had finished this test, but forgot about foundry's `testFail` syntax. Thanks Victor! I've written a new test called `testQuorumNotReached`

examples/governance/methods/guest/Cargo.toml
sashaaldrick: I'm not super familiar with this, so please correct this:

adding a line with [workspace] in `methods/guest/Cargo.toml` would declare this as the root of a cargo workspace.

In this case, I wouldn't be able to use workspace dependencies from the root Cargo.toml i.e. `governance/Cargo.toml`. Seeing as `apps/Cargo.toml` also using workspace dependencies, it seems like its best to keep it the way it is.

Is this correct?

sashaaldrick: I will add the ECDSA acceleration next.

examples/governance/apps/README.md
```diff
@@ -0,0 +1,88 @@
...
+  -V, --version
+          Print version
+```
+
+## Vote Data Tools
+
+This repository contains two Rust scripts for handling voting-related operations:
+
+1. `generate_vote_data.rs`: Generates vote data for a proposal.
+2. `generate_proposal_id.rs`: Generates a proposal ID.
```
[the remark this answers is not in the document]
sashaaldrick: after discussion with @nategraf, this is scoped for a future PR. The current next priority on the governance example is to add Steel, and after that is done, we can pick this up again

examples/governance/methods/guest/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+  "std",
+  "ecdsa",
+], default-features = false }
+risc0-zkvm = { git = "https://github.com/risc0/risc0", branch = "main", default-features = false, features = ["std"] }
+tiny-keccak = { version = "2.0", features = ["keccak"] }
+
+[patch.crates-io]
+# Placing these patch statement in the workspace Cargo.toml will add RISC Zero SHA-256 and bigint
+# multiplication accelerator support for all downstream usages of the following crates.
+sha2 = { git = "https://github.com/risc0/RustCrypto-hashes", tag = "sha2-v0.10.6-risczero.0" }
```
nategraf, line 26:
```suggestion
sha2 = { git = "https://github.com/risc0/RustCrypto-hashes", tag = "sha2-v0.10.8-risczero.0" }
```

examples/governance/methods/guest/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+k256 = { git = "https://github.com/risc0/RustCrypto-elliptic-curves", tag = "k256/v0.13.1-risczero.1" }
+crypto-bigint = { git = "https://github.com/risc0/RustCrypto-crypto-bigint", tag = "v0.5.2-risczero.0" }
```
nategraf, line 28:
```suggestion
crypto-bigint = { git = "https://github.com/risc0/RustCrypto-crypto-bigint", tag = "v0.5.5-risczero.0" }
```

examples/governance/contracts/test/Elf.sol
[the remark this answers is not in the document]
sashaaldrick: My bad, the path is actually correct, but I had changed the path of where the Elf.sol is built to be consistent with the other examples and forgot to remove the git cache for this one.

nategraf (approved):
Looks good! Thanks for continuing to chip away at this one!

PR #198, opened 2024-08-19
WEB3-80: Drop `no-default-features` when installing `cargo-risczero`

capossele:
This should fix:

```
error[E0599]: no method named `env` found for struct `Arg` in the current scope
  --> risc0/cargo-risczero/src/commands/datasheet/mod.rs:41:38
   |
41 |     #[arg(long, value_name = "PATH", env = "RISC0_SERVER_PATH")]
   |                                      ^^^ method not found in `Arg`
```

Note that this behaviour started after merging https://github.com/risc0/risc0/pull/2196, that introduced the `env` usage in claps without adding the `env` feature

### DIFF
--- a/.github/actions/cargo-risczero-install/action.yaml
+++ b/.github/actions/cargo-risczero-install/action.yaml
@@ -28,7 +28,7 @@ runs:
           ref: ${{ inputs.ref }}
           lfs: true
       - name: install cargo-risczero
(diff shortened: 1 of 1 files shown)


nategraf:
We should be sending this issue to the `risc0` repo instead of changing behavior here.

capossele:
Alright, I moved that in `risc0`: https://github.com/risc0/risc0/pull/2254

Closing here for the moment

PR #204, opened 2024-08-24
Derive Clone for EvmEnv

intoverflow:


Wollac (commented):
With https://github.com/risc0/risc0-ethereum/pull/91 we've introduced the ability to execute multiple calls using the same `EvmEnv`, so this feature would only be useful in a case where multiple independent Steel proofs (but still referencing the same block) are to be generated within the same host file.
I am a bit reluctant to make the env clonable, as this would also clone the underlying DB with all its current entries. This could lead to errors or misunderstandings as to whether the cloned envs have the same state or not. 
If this is really a realistic use case, are there perhaps simple alternatives? Like maybe introducing an EnvBuilder?

intoverflow:
You make some very good points. I will think some more and get back to you.

PR #215, opened 2024-09-02
WEB3-99: feat: add EvmEnvBuilder

Wollac:
Add the `EvmEnvBuilder` to simplify the creation of an `EvmEnv` on the host.
See #204 for more context.


steel/src/host/mod.rs
```diff
@@ -45,8 +45,11 @@ pub(crate) type HostEvmEnv<D, H> = EvmEnv<TraceDb<D>, H>;
 impl EthEvmEnv<TraceDb<AlloyDb<Http<Client>, Ethereum, RootProvider<Http<Client>>>>> {
     /// Creates a new provable [EvmEnv] for Ethereum from an HTTP RPC endpoint.
     pub async fn from_rpc(url: Url, number: BlockNumberOrTag) -> Result<Self> {
```
Wollac, line 47: Should this now redundant function be removed, or at least marked as `#[deprecated]`?

steel/src/host/mod.rs
```diff
@@ -60,20 +63,11 @@ where
 {
     /// Creates a new provable [EvmEnv] from an alloy [Provider].
     pub async fn from_provider(provider: P, number: BlockNumberOrTag) -> Result<Self> {
```
Wollac, line 65: Should this now redundant function be removed, or at least marked as `#[deprecated]`?

Wollac:
supersedes #204

capossele (approved):

steel/src/host/mod.rs
```diff
@@ -103,3 +97,96 @@ where
...
+/// Builder to construct an [EvmEnv] on the host.
+#[derive(Clone, Debug)]
+pub struct EvmEnvBuilder<P, H> {
+    provider: P,
+    block: BlockNumberOrTag,
+    phantom: PhantomData<H>,
+}
+
+/// First stage of the [EvmEnvBuilder] without a specified [Provider].
+pub struct NoProvider {}
```
nategraf, line 117: Using the unit-like definition means that there is a single instantiation of the struct, which I think is more semantically aligned here.
```suggestion
pub struct NoProvider;
```

nategraf (approved):

steel/src/host/mod.rs
Wollac: fixed in 4affaae15a32066e132832ddd647ce84c9152969

PR #219, opened 2024-09-04
feat: configurable `eth_getProof` chunk size

Wollac:
Makes the EIP-1186 proof request chunk size configurable in the builder.

closes WEB3-56

### DIFF
--- a/steel/src/host/db/provider.rs
+++ b/steel/src/host/db/provider.rs
@@ -21,6 +21,7 @@ use anyhow::{ensure, Result};
 use revm::Database;
 
 #[derive(Clone, Debug)]
+#[non_exhaustive]
 pub struct ProviderConfig {
     /// Max number of storage keys to request in a single `eth_getProof` call.
     pub eip1186_proof_chunk_size: usize,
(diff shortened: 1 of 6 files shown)


steel/src/host/db/provider.rs
```diff
@@ -20,15 +20,29 @@ use alloy_primitives::{Address, BlockNumber, StorageKey};
 use anyhow::{ensure, Result};
 use revm::Database;
 
+#[derive(Clone, Debug)]
```
nategraf, line 23: I imagine we'll want to add fields to this in the future, so marking it non-exhaustive would make sense. 
```suggestion
#[derive(Clone, Debug)]
#[non_exhaustive]
```
Constructing it then means you need a builder pattern (or the dev can get the default and modify it as `mut`). One way to do that is the self-contained version like we use for `ProverOpts`, where each field has a `with_field` method that consumes `self` and returns the modified version.
https://github.com/risc0/risc0/blob/main/risc0/zkvm/src/host/client/prove/mod.rs#L288

nategraf (approved):

steel/src/host/db/provider.rs
Wollac: I agree, marking it `non_exhaustive` is definitely better.
Currently, since it has only one field, it can be set directly in the EnvBuilder: https://github.com/risc0/risc0-ethereum/blob/fdf72bb848d987cb7fc4050f25bfafec0d88c28e/steel/src/host/mod.rs#L168

But once we add more fields, I like the patter you mentioned.

PR #231, opened 2024-09-10
WEB3-115: chore: Add more Steel precompile tests

Wollac:
This PR adds tests against the precompiles behind `revm` feature flags. This includes the Cancun point evaluation precompile, which is currently disabled in Steel due to performance concerns with `c-kzg`.


steel/tests/steel.rs
```diff
@@ -143,6 +176,32 @@ async fn precompile() {
     );
 }
 
+// The EIP-4844 point evaluation precompile has been disabled due to performance concerns in the zkVM.
```
nategraf, line 179: I've opened an issue for this. It would be good to have a pass through feature flag, e.g. to enable `kzg-rs` or `c-kzg` on `revm`, even if performance is not where we want to be. If performance if bad enough to impact UX, turning it off by default makes sense, but we also want a switch to turn it on in case a use case is blocked by it and willing to pay the cost.
```suggestion
// TODO(#237): EIP-4844 point evaluation precompile is not currently enabled.
```

nategraf (approved):

PR #236, opened 2024-09-11
WEB3-119: vendor git dependencies

Wollac:
This PR adds the following unpublished crate as a vendored module to Steel:
- https://github.com/ralexstokes/ethereum-consensus/tree/cf3c404043230559660810bc0c9d6d5a8498d819

### DIFF
--- a/steel/build.rs
+++ b/steel/build.rs
@@ -21,8 +21,9 @@ use std::{
 };
 use walkdir::WalkDir;
 
-const SOURCE: &'static str = "src/vendor/git/ethereum-consensus/ethereum-consensus/src";
-const TARGET: &'static str = "src/vendor/ethereum_consensus";
+const SOURCE: &str = "src/vendor/git/ethereum-consensus/ethereum-consensus/src";
+const TARGET: &str = "src/vendor/ethereum_consensus";
+
 fn main() -> Result<()> {
     let manifest_dir =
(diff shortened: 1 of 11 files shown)


steel/build.rs
```diff
@@ -0,0 +1,52 @@
...
+    let mut buffer = [0u8; MAX_MERKLE_TREE_DEPTH * BYTES_PER_CHUNK];
+    for i in 0..MAX_MERKLE_TREE_DEPTH - 1 {
+        let focus_range = i * BYTES_PER_CHUNK..(i + 2) * BYTES_PER_CHUNK;
+        let focus = &mut buffer[focus_range];
+        let (source, target) = focus.split_at_mut(BYTES_PER_CHUNK);
+        hash_nodes(&mut hasher, source, source, target);
+    }
+    buffer
+}
+
```
nategraf, line 25: Would this comment be accurate?
```suggestion
NOTE: This build script is required by the vendored dependency ssz-rs.
```

nategraf (commented):
This is looking good.

Could you add a note, perhaps in a `vendor/README.md`, on what commit these came from? I'd like for myself or anyone else to be able to confirm that these match those sources by recopying the sources from GitHub and confirming that they match.

nategraf:
Closing this PR for now. We may revisit this in the future.

PR #239, opened 2024-09-11
WEB3-107: fix: Specific error when `Contract::preflight` was missed.

### LINKED ISSUE #224: Forgetting to run `Contract::preflight` results in difficult to debug error
An application developer using Steel ran into an issue when using Steel when they forgot to run `Contract::preflight` on the host. It is expected that this will result in a failure. However the following assert was what caught the issue, and they had a very difficult time debugging it.

https://github.com/risc0/risc0-ethereum/blob/8dc728708a866e9b1a1bf7cb83f507c1929199b4/steel/src/block.rs#L35

It is likely this will be a common error, and so we should make it easier to figure out how to fix it.

Wollac:
Returns "no accounts accessed: use Contract::preflight" when no `Contract::preflight` was called.

closes #224 
closes WEB3-107


nategraf (approved):
Nice, I think this will be helpful

PR #241, opened 2024-09-12
WEB3-105: chore: corruption tests for Steel

### LINKED ISSUE #222: Corruption tests for Steel
Add tests to Steel that ensure any corruption of the data will be detected by the zkVM guest, or during verification against a the set of trusted block commitments. This serves as a basic check against unconstrained modification of the Steel query results, complimenting our tests of the "happy path".

Examples of data we need to check:
* Modification of a value in EVM storage.
* Modification or omission of nodes in the storage MPT inclusion proof.
* Modification of the contract code.
* Modification of account balances.
* Modification or omission of nodes in the state MPT inclusion proof.
* Modification of the execution block header.
* Modification of the consensus block header.

Wollac:
This PR adds tests to ensure that corruptions of the input leads to errors. For this, the tests modify the JSON representation of the `EvmInput` directly.

It checks the following corruptions:
- modification of a single storage value
- removing parts of a storage trie
- modification of an account's balance
- removing parts of the state trie
- modification of a single contract's code
- modification of a header in the ancestor chain
- modification of the execution block header
- modification of a single element in the beacon commitment Merkle path

closes #222 
closes WEB3-105


nategraf (approved):
This is great! Thanks for implementing these

PR #243, opened 2024-09-12
WEB3-124: chore: address TODOs

Wollac:
This PR either resolves the open TODOs, or links them to an GitHub issue.


steel/src/ethereum.rs
```diff
@@ -94,6 +94,7 @@ impl EvmBlockHeader for EthBlockHeader {
         // technically, this is only valid after EIP-4399 but revm makes sure it is not used before
         blk_env.prevrandao = Some(header.mix_hash);
         if let Some(excess_blob_gas) = header.excess_blob_gas {
+            // safe conversion: according to the consensus spec, excess_blob_gas is always an uint64
             blk_env.set_blob_excess_gas_and_price(excess_blob_gas.try_into().unwrap())
```
nategraf, line 98: Annoying that the upstream header type is factored to use ` u128`. I worry that an `unwrap` here would become a DoS vector, in that a malformed header could crash the host. I think we should use `as u64` instead, to simple truncate the result.
```suggestion
            // safe conversion: according to the consensus spec, excess_blob_gas is always an uint64
            blk_env.set_blob_excess_gas_and_price(excess_blob_gas as u64)
```

nategraf (approved):

steel/src/contract.rs
```diff
@@ -286,13 +289,15 @@ where
...
+        // safe unwrap: env is never returned without a DB
         let state_db = self.env.db.as_ref().unwrap();
         let mut evm = new_evm::<_, H>(
             WrapStateDb::new(state_db),
             self.env.cfg_env.clone(),
             self.env.header.inner(),
         );
-        self.tx.transact(&mut evm).unwrap()
+        // typically, guest functions should panic instead of returning an error
+        self.tx.transact(&mut evm).expect("call failed")
```
nategraf, line 300: I wouldn't say this is true. We do use panic liberally in guest code, and in the `main` especially, but guest libraries can use errors instead, and generally should, I think, as that they can optionally be handled by the caller.

I think in this case, I would return the result produced by `self.tx.transact`

nategraf, line 300:
```suggestion
        // typically, guest functions should panic instead of returning an error
        self.tx.transact(&mut evm)
```

steel/src/contract.rs
```diff
@@ -286,13 +289,15 @@ where
     ///
     /// [EvmEnv]: crate::EvmEnv
     pub fn call(self) -> C::Return {
```
nategraf, line 291: Side note that it'd be good to use an error type that (when `std` is enabled) implements `std::error::Error`. We can address that later though (or not at all. It's not critical)
```suggestion
    pub fn call(self) -> Result<C::Return, String> {
```

nategraf (approved):

steel/src/ethereum.rs
Wollac: I'm not sure, isn't a weird overflow even worse than a clean panic?
If I am a malicious host/RPC, there are so many ways I could make the guest panic, like sending wrong state root or corrupted inclusion proofs, ...

steel/src/contract.rs
Wollac: OK, that was my impression that we often prefer panicking over errors in the guest such as in `env::read()` or other `env`s.

steel/src/contract.rs
Wollac: 100% agree, after all it should be a `thiserror` or something. However, it was not that trivial to do, so I decided to leave it as a string for now.

steel/src/contract.rs
Wollac: I have changed that in 17b4550b97a91037dbef2d1cadc171fd6b5f5594

nategraf: It depends, ya. I'd say as a rule of thumb that guest panicking when the error could only have been caused by host misbehavior, ya (e.g. in the `k256` patch, I added a panic if the host did not normalize the result). In other cases, it's a judgement call. I lean more towards not panicking anywhere except the root of the guest (i.e. in `main.rs`)

PR #251, opened 2024-09-24
WEB3-136: make more Steel types public

Wollac:
This PR makes the following types public:
- `BeaconInput`
- `BlockInput`
- `StateDb`

It also generalizes the `BeaconInput` concept of having a `BlockInput` and a header-based commitment.


Wollac:
Make it a draft for now to be sure that we don't need to expose even more

capossele (approved):

PR #272, opened 2024-10-02
WEB3-151: Use builder also for beacon input

### LINKED ISSUE #259: Set the beacon API URL in the builder
To enable a clear separation (in the Builder) between state (query) block and commitment block, move the URL from `into_beacon_input` into the builder.

Wollac:
Previously, an input with a Beacon root commitment was created using the `into_beacon_input` method of the environment. This is inconsistent as all configuration should happen in the `EvmEnvBuilder`. 

Resolves #259 
Resolves WEB3-151

### DIFF
--- a/steel/src/block.rs
+++ b/steel/src/block.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{state::StateDb, EvmBlockHeader, EvmEnv, GuestEvmEnv, MerkleTrie};
+use crate::{state::StateDb, Commitment, EvmBlockHeader, EvmEnv, GuestEvmEnv, MerkleTrie};
 use ::serde::{Deserialize, Serialize};
 use alloy_primitives::{map::HashMap, Bytes};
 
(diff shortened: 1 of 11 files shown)


steel/src/block.rs
```diff
@@ -27,6 +27,11 @@ pub struct BlockInput<H> {
 }
 
 impl<H: EvmBlockHeader> BlockInput<H> {
+    /// Returns the block header this input is based on.
+    pub fn header(&self) -> &H {
```
Wollac, line 31: Is this really used?

steel/src/host/mod.rs
```diff
@@ -115,165 +117,40 @@ where
...
-            sealed_header.seal(),
-        ));
+impl<T, P> EthHostEvmEnv<AlloyDb<T, Ethereum, P>, ()>
+where
+    T: Transport + Clone,
+    P: Provider<T, Ethereum>,
+{
+    /// Converts the environment into a [EvmInput] committing to an Ethereum Beacon block root.
+    #[deprecated(
+        since = "0.13.0",
```
Wollac, line 146: Fix version

Wollac:
To be merged after https://github.com/risc0/risc0-ethereum/pull/265

capossele (approved):

nategraf (approved):

PR #275, opened 2024-10-03
WEB3-164: added new Steel readme + relevant docs

sashaaldrick:
This cleans up the README to add a new introduction, documentation for a Steel walkthrough, and getting started with Steel using `create-steel-app` for erc20-counter.


steel/docs/steel-commitments.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+This block hash has to be compared on-chain, alongside the verification of the proof, to validate that the Steel proof is correct and to verify the integrity of RPC data.
+
+## What is a Steel Commitment?
+
+A commitment consists of two values: the block ID and the block digest.  The block ID encodes two values, the Steel version number and a block identifier in an encoding which is effectively: `abi.encodePacked(uint16 version, uint240 blockIdentifier)`.
+
+```solidity
+struct Commitment {
```
Wollac, line 61: Needs to be adapted according to https://github.com/risc0/risc0-ethereum/pull/281

steel/docs/steel-commitments.md
```diff
@@ -0,0 +1,138 @@
...
+	if (block.timestamp - blockTimestamp > 12 * 8191) {
+		revert CommitmentTooOld();
+	}
+	return blockRoot == Beacon.blockRoot(blockTimestamp);
+}
+```
+
+The second method allows validation using the [EIP-4788] beacon roots contract. This technique extends the time window in which the proof can be validated on-chain to just over a day, making it suitable for scenarios requiring more extensive computation. It requires access to a beacon API endpoint and can be enabled by calling `EvmEnv::into_beacon_input`. However, this approach is specific to Ethereum Steel proofs and depends on the implementation of EIP-4788.
+
+Note that EIP-4788 only provides access to the parent beacon root, requiring iterative queries in Solidity to retrieve the target beacon root for validation. This iterative process can result in slightly higher gas costs compared to using the `blockhash` opcode. Overall, it is suitable for environments where longer proof generation times are required.
```
Wollac, line 126: Needs to be adapted according to https://github.com/risc0/risc0-ethereum/pull/265

steel/docs/how-does-steel-work.md
```diff
@@ -0,0 +1,130 @@
...
+let account: Address = env::read();
+
+let evm_env = input.into_env().with_chain_spec(&ETH_SEPOLIA_CHAIN_SPEC);
+
+// Execute the view call; it returns the result in the type generated by the `sol!` macro.
+let call = IERC20::balanceOfCall { account };
+let returns = Contract::new(contract, &evm_env)
+	.call_builder(&call)
+	.call()
+	.unwrap();
```
Wollac, line 30:
```suggestion
	.call();
```

steel/README.md
```diff
@@ -1,150 +1,56 @@
...
-    println!("View call result: {}", returns._0);
-}
-```
+The recommended place is to start is [Steel examples], specifically the [ERC20 Counter] example. 
 
-## Host code
+The [create-steel-app] script will allow you to set up the erc20-counter example locally in one command:
 
-Here is a snippet to the [relevant code](../examples/erc20/host/src/main.rs) on the host, it requires the same arguments as the guest:
+`sh -c "$(curl -fsSL https://raw.githubusercontent.com/risc0/risc0-ethereum/refs/heads/main/steel/docs/create-steel-app/create-steel-app)"`
```
nategraf, line 17: It's better to use file-io based execution rather than args. We can't use pipe `|` since we need `stdin` (I assume this is why we are using `-c` here) but there is another way.
````suggestion
```
sh <(curl -fsSL https://raw.githubusercontent.com/risc0/risc0-ethereum/refs/heads/main/steel/docs/create-steel-app/create-steel-app)
```
````

File: steel/docs/create-steel-app/README.md [hunk 5]
```diff
@@ -0,0 +1,50 @@
...
+
+Make sure to have the following installed:
+
+1. [Rust]
+2. [Foundry]
+3. [cargo-risczero]
+
+## Usage
+
+`sh -c "$(curl -fsSL https://raw.githubusercontent.com/risc0/risc0-ethereum/refs/heads/main/steel/docs/create-steel-app/create-steel-app)"`
```
nategraf, line 17:
````suggestion
```
sh <(curl -fsSL https://raw.githubusercontent.com/risc0/risc0-ethereum/refs/heads/main/steel/docs/create-steel-app/create-steel-app)
```
````

File: steel/docs/create-steel-app/README.md [hunk 6]
```diff
@@ -0,0 +1,50 @@
...
+
+Once the script is finished running, you should:
+
+```
+cd PROJECT_NAME
+cargo build
+forge build
+``` 
+
+After this, you can export the necessary Bonsai environment variables:
```
nategraf, line 29:
```suggestion
After this, you can export the necessary Bonsai environment variables if you'd like to use Bonsai. Otherwise, local proving will be used:
```

steel/docs/create-steel-app/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+[create-steel-app]: create-steel-app
+[erc20-counter]: https://github.com/risc0/risc0-ethereum/tree/main/examples/erc20-counter
+[documentation]: ../../README.md#documentation
+[Rust]: https://www.rust-lang.org/tools/install
+[Foundry]: https://book.getfoundry.sh/getting-started/installation
+[cargo-risczero]: https://dev.risczero.com/api/zkvm/install
+[1.0]: https://github.com/risc0/risc0-ethereum/tree/release-1.0
+[1.1]: https://github.com/risc0/risc0-ethereum/tree/release-1.1
+
```
nategraf, line 50:
```suggestion
```

nategraf (commented):
As of right now, `forge build` with the new `create-steel-app` fails

```
Compiler run failed:
Error (9582): Member "blockRoot" not found or not visible after argument-dependent lookup in type(library Beacon).
  --> contracts/test/Counter.t.sol:76:30:
   |
76 |         bytes32 beaconRoot = Beacon.blockRoot(beaconTimestamp);
   |                              ^^^^^^^^^^^^^^^^

Error:
Compilation failed
```

steel/docs/how-does-steel-work.md
```diff
@@ -0,0 +1,130 @@
...
+        // Verify the proof
+        bytes32 journalHash = sha256(journalData);
+        verifier.verify(seal, imageID, journalHash);
+        
+	 // If the balance is at least one, update the counter
+        counter += 1;
+    }
+```
+
+Within a single proof, we’ve seen Steel can handle hundreds of thousands of view calls. Specifically, one partner application has shown gas savings of 1.2 *billion* gas for a contract call using around 400,000 SLOADs. 1.2 billion gas is around 30 *blocks* worth of execution and this can be verified onchain in one proof, that costs under $10 to generate, and less than 300k gas to verify (see [RISC Zero’s verification contracts]).
```
nategraf, line 112: Hundreds of thousands of view calls is slightly different than what we've seen. It should be possible, but is not something we've actually done yet.
```suggestion
Within a single proof, we’ve seen Steel can handle view calls orders of magnitude larger than on-chain execution can handle. Specifically, one partner application has shown gas savings of 1.2 *billion* gas for a contract call using around 400,000 SLOADs. 1.2 billion gas is around 30 *blocks* worth of execution and this can be verified onchain in one proof, that costs under $10 to generate, and less than 300k gas to verify (see [RISC Zero’s verification contracts]).
```

steel/docs/how-does-steel-work.md
```diff
@@ -0,0 +1,130 @@
...
+
+With proof aggregation, cost savings are amortized even further, by taking multiple separate applications of Steel, and wrapping them all up into a single SNARK.
+
```
nategraf, line 115:
```suggestion
With proof aggregation, cost savings are amortized even further, by taking multiple separate applications of RISC Zero, and wrapping them all up into a single SNARK.
Aggregation is a key feature of [Boundless][boundless-website]

[boundless-website]: https://beboundless.xyz/

```

steel/docs/steel-commitments.md
```diff
@@ -0,0 +1,138 @@
...
+
+This block is used for RPC queries (i.e. `getStorageAt`) during the `preflight` call. Once the preflighting is done, `into_input` is called:
+
+```rust
+let evm_input = env.into_input().await?
+```
+
+During this `into_input` step, the preflight data is packed into a form that can be read and validated in the guest.
+
+Crucially, during the step, Steel will use the RPC call [eth\_getProof] for all accounts accessed during the preflight and the return data is combined into a sparse Merkle trie.
```
nategraf, line 28: This link is broken

steel/docs/steel-commitments.md
```diff
@@ -0,0 +1,138 @@
...
+	tokenAddress: contract,
+};
+env::commit_slice(&journal.abi_encode());
+```
+
+This block hash has to be compared on-chain, alongside the verification of the proof, to validate that the Steel proof is correct and to verify the integrity of RPC data.
+
+## What is a Steel Commitment?
+
+A commitment consists of two values: the block ID and the block digest.  The block ID encodes two values, the Steel version number and a block identifier in an encoding which is effectively: `abi.encodePacked(uint16 version, uint240 blockIdentifier)`.
```
nategraf, line 58: If they need this level of detail, they can read the source code. Including it inline is prone to bitrot.
```suggestion
A commitment consists of two values: the block ID and the block digest.  The block ID encodes two values, the Steel version number and a block identifier (e.g. a block number).
```

steel/docs/steel-commitments.md
```diff
@@ -0,0 +1,138 @@
...
+
+```solidity
+struct Commitment {
+	uint256 blockID;
+	bytes32 blockDigest;
+}
+```
+
+In the `increment` function in the `Counter` contract, we saw this require statement: \`\`
```
nategraf, line 67:
```suggestion
In the `increment` function in the `Counter` contract, we saw this require statement:
```

steel/docs/steel-commitments.md
```diff
@@ -0,0 +1,138 @@
...
+
+```solidity
+require(Steel.validateCommitment(journal.commitment), "Invalid commitment");
+```
+
+The [Steel library](https://github.com/risc0/risc0-ethereum/blob/main/contracts/src/steel/Steel.sol) contains the function `validateCommitment` :
+
+```javascript
```
nategraf, line 75:
````suggestion
```solidity
````

File: steel/docs/steel-commitments.md [hunk 14]
```diff
@@ -0,0 +1,138 @@
...
+/// @return True if the block's block hash matches the block hash, false otherwise.
+function validateBlockCommitment(uint256 blockNumber, bytes32 blockHash) internal view returns (bool) {
+	if (block.number - blockNumber > 256) {
+		revert CommitmentTooOld();
+	}
+	return blockHash == blockhash(blockNumber);
+}
+```
+
+This method uses the `blockhash` opcode to commit to a block hash that is no more than 256 blocks old. With Ethereum's 12-second block time, this provides a window of about 50 minutes to generate the proof and ensure that the validating transaction is contained in a block. This approach is ideal for most scenarios, including complex computations, as it typically provides sufficient time to generate the proof.
```
nategraf, line 107: We are actually going to recommend the beacon hash on any chain that its available.
```suggestion
This method uses the `blockhash` opcode to commit to a block hash that is no more than 256 blocks old. With Ethereum's 12-second block time, this provides a window of about 50 minutes to generate the proof and ensure that the validating transaction is contained in a block. This approach will work for most scenarios, including complex computations, as it typically provides sufficient time to generate the proof.
```

steel/docs/steel-commitments.md
```diff
@@ -0,0 +1,138 @@
...
+/// @return True if the block's block root matches the block root, false otherwise.
+function validateBeaconCommitment(uint256 blockTimestamp, bytes32 blockRoot) internal view returns (bool) {
+	if (block.timestamp - blockTimestamp > 12 * 8191) {
+		revert CommitmentTooOld();
+	}
+	return blockRoot == Beacon.blockRoot(blockTimestamp);
+}
+```
+
+The second method allows validation using the [EIP-4788] beacon roots contract. This technique extends the time window in which the proof can be validated on-chain to just over a day, making it suitable for scenarios requiring more extensive computation. It requires access to a beacon API endpoint and can be enabled by calling `EvmEnv::into_beacon_input`. However, this approach is specific to Ethereum Steel proofs and depends on the implementation of EIP-4788.
```
nategraf, line 124: It's not so much about the amount of computation, since with Bonsai and similar we can get a lot of computation proven within the 50 minute window (we've demonstrated proving ~50B cycles in ~30 minutes). But there are other reasons, and folks who need a longer window will know it.
```suggestion
The second method allows validation using the [EIP-4788] beacon roots contract. This technique extends the time window in which the proof can be validated on-chain to just over a day. It requires access to a beacon API endpoint and can be enabled by calling `EvmEnv::into_beacon_input`. However, this approach is specific to Ethereum (L1) Steel proofs and depends on the implementation of EIP-4788.
```

steel/README.md
nategraf: We need a note in `RELEASE.md` to update this based on the branch as part of the release process.

steel/docs/create-steel-app/create-steel-app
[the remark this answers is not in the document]
nategraf: Restricted to 1.1 in 3050beff60ebe43e65566ec9853be1d869299f06

steel/docs/steel-commitments.md
```diff
@@ -0,0 +1,138 @@
...
+/// @notice Validates if the provided beacon commitment matches the block root of the given timestamp.
+/// @param blockTimestamp The timestamp to compare against.
+/// @param blockRoot The block root to validate.
+/// @return True if the block's block root matches the block root, false otherwise.
+function validateBeaconCommitment(uint256 blockTimestamp, bytes32 blockRoot) internal view returns (bool) {
+    if (block.timestamp - blockTimestamp > 12 * 8191) {
+        revert CommitmentTooOld();
+    }
+    return blockRoot == Beacon.blockRoot(blockTimestamp);
+}
```
Wollac, line 121:
```suggestion
/// @notice Validates if the provided block commitment matches the block hash of the given block number.
/// @param blockNumber The block number to compare against.
/// @param blockHash The block hash to validate.
/// @return True if the block's block hash matches the block hash, false otherwise.
function validateBlockCommitment(uint256 blockNumber, bytes32 blockHash) internal view returns (bool) {
    if (block.number - blockNumber > 256) {
        revert CommitmentTooOld();
    }
    return blockHash == blockhash(blockNumber);
}
```

steel/docs/steel-commitments.md
```diff
@@ -0,0 +1,138 @@
...
+```
+
+The second method allows validation using the [EIP-4788] beacon roots contract. This technique extends the time window in which the proof can be validated on-chain to just over a day. It requires access to a beacon API endpoint and can be enabled by calling `EvmEnv::into_beacon_input`. However, this approach is specific to Ethereum (L1) Steel proofs and depends on the implementation of EIP-4788.
```
Wollac, line 124:
```suggestion
The second method allows validation using the [EIP-4788] beacon roots contract. This technique extends the time window in which the proof can be validated on-chain to just over a day. It requires access to a beacon API endpoint and can be enabled by calling `EvmEnv::builder().beacon_api()`. However, this approach is specific to Ethereum (L1) Steel proofs and depends on the implementation of EIP-4788.
```

PR #281, opened 2024-10-04
WEB3-166: Commit Chain Configuration to Journal

Wollac:
Currently, Steel allows flexible chain configurations but doesn't commit them to the journal. This can lead to:
- Accidental execution on the wrong chain (e.g., Sepolia spec on mainnet).
- Difficulty in verifying the integrity of custom configurations.

This PR resolves these issues by hashing the chain configuration and including it in the Steel `Commitment`. This ensures transparency and prevents potential security risks.


capossele (approved):

steel/src/config.rs
```diff
@@ -86,6 +101,55 @@ impl ChainSpec {
     }
 }
 
+/// A simple structured hasher.
```
nategraf, line 104:
```suggestion
// NOTE: We do not want to make this public, to avoid having multiple traits with the `digest` function in
// the RISC Zero ecosystem of crates.
/// A simple structured hasher.
```

nategraf (approved):

PR #284, opened 2024-10-07
Cherry-pick main and set version to 1.1.4

Wollac:
Cherry picks all changes to Steel since the last release of Steel. Updates all versions to 1.1.4, including Steel.

**This PR marks the 1.0 release of Steel**


nategraf:
The `CI / examples` job failed on the last test, which was deprecated but covered in other jobs.
https://github.com/risc0/risc0-ethereum/actions/runs/11218972638/job/31183902249?pr=284

PR #290, opened 2024-10-11
WEB3-172: Add SECURITY.md with information about security policy and audits

nategraf:


kevinnassery:
Checked this, looks good.  Note it sets a baseline that we'll put out advisories in multiple repos; which is probably best but will require folks to subscribe to them if they want alerts.

Wollac (approved):

PR #298, opened 2024-10-18
WEB3-179: feat: Steel history

Wollac:
This PR introduces a new mechanism for verifying historical commitments, decoupling EVM execution blocks from commitment blocks. This is achieved by:

* **Iterative Historical State Verification:**  Starting with a recent state ($S_0$) confirmed by the canonical chain, the Beacon Roots contract is used to verify the state 24 hours prior ($S_1$). This process is repeated, stepping back 24 hours at a time,  until the desired historical state ($S_n$) is reached.

* **Security:** By anchoring the verification process to the canonical chain ($S_0$), we ensure that the historical state ($S_n$) used for contract execution is valid and tamper-proof.

This enhancement enables more flexible and efficient handling of historical data within Steel.

Closes WEB3-179
Closes WEB3-202


steel/src/history/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+/// Input committing a previous block hash to the corresponding Beacon Chain block root.
+pub type HistoryInput<H> = ComposeInput<H, HistoryCommit>;
+
+/// Verifiable proof that an execution block hash is included in the past of a specific beacon block
+/// on the Ethereum blockchain.
+///
+/// This struct encapsulates the necessary data to prove that a given execution block is part of the
+/// canonical chain according to the Beacon Chain.
+#[derive(Clone, Serialize, Deserialize)]
+pub struct HistoryCommit {
```
nategraf, line 33: How could a guest constrain the block it runs it's queries against (tied to the `evm_commit`)? As it stands I don't see any way for the guest to e.g. check the block number or hash such that it could either place a constraint on it, or commit it to the journal to be checked in the application contract.

steel/src/history/mod.rs
```diff
@@ -0,0 +1,254 @@
...
+            evm_header: &Sealed<EthBlockHeader>,
+            commitment_header: &Sealed<EthBlockHeader>,
+            rpc_provider: P,
+            beacon_url: Url,
+        ) -> anyhow::Result<Self>
+        where
+            T: Transport + Clone,
+            P: Provider<T, Ethereum>,
+        {
+            let client = BeaconClient::new(beacon_url.clone()).context("invalid URL")?;
```
nategraf, line 111: We might want to start with a check that `evm_header` is at least as old as `commitment_header`.

steel/src/host/builder.rs
```diff
@@ -202,14 +203,38 @@ impl<P, H> EvmEnvBuilder<P, H, ()> {
...
 
+/// Config for separating the execution block from the commitment block.
+#[derive(Clone, Debug)]
+pub struct History {
+    beacon_url: Url,
+    commitment_block: BlockNumberOrTag,
+}
+
 impl<P> EvmEnvBuilder<P, EthBlockHeader, Url> {
+    /// Sets a dedicated block for the commitment that is different from the execution block.
```
nategraf, line 214:
```suggestion
    /// Sets a dedicated block for the commitment that is different from the execution block.
    ///
    /// The commitment block must be later than the execution block (i.e. the execution block must be
    /// an ancestor of the commitment block). This allows executing smart contracts with historical
    /// state (e.g. 30 days ago), and verifying the results against a recent block commitment.
    ///
    /// Note that this feature requires a Beacon chain RPC provider, as it uses EIP-4788.
```

steel/src/host/builder.rs
```diff
@@ -230,11 +255,56 @@ impl<P> EvmEnvBuilder<P, EthBlockHeader, Url> {
     }
 }
 
+impl<P> EvmEnvBuilder<P, EthBlockHeader, History> {
```
nategraf, line 258: This type-state generic builder pattern is really nice!

nategraf (commented):
This looks great! I have a couple of questions looking to getting this released:

1. No tests currently cover the new functionality in CI. I know we need a consensus RPC to run the tests, which is a challenge for doing so frequently. One option might be to have a CI test that is required, but does not run on every commit. Searching for ways to do this, it's not too hard to set up a trigger to run it when a comment with`/run-steel-tests` is posted. This might be a little annoying, since that'll mean this command needs to be sent every time the tests needs to be run, but we could try it out and see if we think of any better ideas.
2. We should have some documentation about this feature and make sure it's easy to find. Probably just start with an update to the README and/or `steel/docs`, as well as a comment in the CHANGELOG of course.
3. Should we mark the history functionality with `unstable`? I lean towards yes. It's a major new feature, and I wouldn't be surprised if we get some feedback that would be best addressed with breaking changes. There is also the matter of security review. We should think about it a bit more, but this new functionality feels subtle enough that getting an audit of the implementation might be a good step before we endorse it for wider use, even though the actual code footprint is pretty small.

   If yes, we can use the `stability` crate that we use in `risc0` and mark the `history` module with `#[unstable(feature = "history")]`. With this, we can get this out in the next release and start to get feedback without committing to the exact API or endorsing it for production just yet.

steel/src/history/beacon_roots.rs
```diff
@@ -0,0 +1,221 @@
...
+        network::BlockResponse,
+        providers::{Provider, ProviderBuilder},
+        rpc::types::BlockNumberOrTag as AlloyBlockNumberOrTag,
+    };
+    use test_log::test;
+
+    const EL_URL: &str = "https://ethereum-rpc.publicnode.com";
+
+    #[test(tokio::test)]
+    #[ignore] // This queries actual RPC nodes, running only on demand.
```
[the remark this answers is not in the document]
Wollac: We have this in several places, let's clean it all up in another PR.

steel/src/history/mod.rs
Wollac: In the guest, the `EthEvmEnv` gives you access to the (EVM query) header. You can do things like:
```rust
// Read the input from the guest environment.
let input: EthEvmInput = env::read();
let evm_env = input.into_env();

let block_hash = evm_env.header().seal();
let block_number = evm_env.header().number;
...
```

This header is exactly what will be verified by a `BlockHeaderCommit`. Do you think this is enough for your use case?

steel/src/history/mod.rs
Wollac: Addressed in 836980a2203abd25e40933d7d1895563f21d1778

steel/src/host/builder.rs
Wollac: It is. Unfortunately, having non-compiletime config changes like
```rust
if let Some(beacon_api_url) = args.beacon_api_url { {
    builder = builder.beacon_api(beacon_api_url);
}
```
Is not possible with this pattern.

Wollac:
> 1. No tests currently cover the new functionality in CI. 

I've created issue #307 to add the history commitment to the CI E2E test.
In addition, we already have two unit tests `history::tests::from_beacon_commit_and_header` and `host::builder::tests::build_history_env` that test this feature. However, since they test against the live Ethereum testnet, they are ignored by default. (We might consider enabling them under certain conditions.)

> 2. We should have some documentation about this feature and make sure it's easy to find. Probably just start with an update to the README and/or `steel/docs`, as well as a comment in the CHANGELOG of course.

CHANGELOG changes have been added to this PR, the missing documentation is documented in #308.

> 3. Should we mark the history functionality with `unstable`? 

I am not sure how this would actually work... The only out-facing API change is adding `.commitment_block()` to the builder everything else including the `history` module is just adding a new variant to an enum which should be transparent to the user.

steel/src/history/mod.rs
nategraf: Ah! That's perfect, thank you

nategraf (approved):

PR #311, opened 2024-10-28
Use `unstable` to mark Steel history commitments as not stable

Wollac:
fixes WEB3-202

### DIFF
--- a/steel/src/lib.rs
+++ b/steel/src/lib.rs
@@ -26,7 +26,10 @@ mod block;
 pub mod config;
 mod contract;
 pub mod ethereum;
+#[cfg(feature = "unstable-history")]
 pub mod history;
+#[cfg(not(feature = "unstable-history"))]
+mod history;
 #[cfg(feature = "host")]
 pub mod host;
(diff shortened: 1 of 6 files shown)


steel/src/lib.rs
```diff
@@ -37,10 +40,14 @@ mod state;
...
 pub use block::BlockInput;
 pub use contract::{CallBuilder, Contract};
-pub use history::HistoryInput;
 pub use mpt::MerkleTrie;
 pub use state::{StateAccount, StateDb};
 
+#[cfg(feature = "unstable-history")]
+pub use history::HistoryInput;
+#[cfg(not(feature = "unstable-history"))]
+pub(crate) use history::HistoryInput;
```
Wollac, line 49: I don't think there is an easy way to make EvmInput::History private, but since no such input can be created by the public API, and since it contains only private fields, this should be fine.

nategraf: Agreed

nategraf (approved):

PR #313, opened 2024-10-29
WEB3-208: Add IRiscZeroSelectable interface

capossele:
Thi PR introduces a new interface called `IRiscZeroSelectable` that leverages the already existing `SELECTOR` constant of the `RiscZeroGroth16Verifier` (and more in general of any base implementation) to abstract from the base implementation when requiring the value of its selector (e.g., during deployment).

It also adds a `Makefile` to quickly deploy a local devnet and  fixes some printing to create a proper TOML block e.g.,

```console
  [[chains.anvil.verifiers]]
  version = "1.2.0-alpha.1"
  selector = "0x50bd1769"
  verifier = "0x9fe46736679d2d9a65f0992f2272de9f3c7fa6e0"
  estop = "0xcf7ed3acca5a467e9e704c703e8d87f634fb0fc9"
```


contracts/test/RiscZeroGroth16Verifier.t.sol
```diff
@@ -138,6 +139,7 @@ contract RiscZeroGroth16VerifierTest is Test {
     }
 
     function testSelectorIsStable() external view {
-        require(verifier.SELECTOR() == hex"50bd1769");
+        IRiscZeroSelectable selectable = IRiscZeroSelectable(verifier);
+        require(selectable.SELECTOR() == hex"50bd1769");
```
nategraf, line 143: I don't think this change should be required, right? Because the `SELECTOR` function exists on `RiscZeroGroth16Verifier`

contracts/src/IRiscZeroSelectable.sol
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+pragma solidity ^0.8.9;
+
+/// @notice Selectable interface for RISC Zero verifier.
+interface IRiscZeroSelectable {
+    /// @notice The selector for the verifier.
```
nategraf, line 21: Adding the longer description here would be nice
```suggestion
    /// @notice A short key attached to the seal to select the correct verifier implementation.
    /// @dev The selector is taken from the hash of the verifier parameters If two
    ///      receipts have different selectors (i.e. different verifier parameters), then it can
    ///      generally be assumed that they need distinct verifier implementations. This is used as
    ///      part of the RISC Zero versioning mechanism.
    ///
    ///      A selector is not intended to be collision resistant, in that it is possible to find
    ///      two preimages that result in the same selector. This is acceptable since it's purpose
    ///      to a route a request among a set of trusted verifiers, and to make errors of sending a
    ///      receipt to a mismatching verifiers easier to debug. It is analogous to the ABI
    ///      function selectors.
```

nategraf:
I like the contents of this PR, except the Makefile, so I removed that file in 44c372b and am going to merge this PR.

My understanding is that the problem that the makefile is solving is that we need a good way to create a deployment that includes the router and associated verifiers. Instead of a `Makefile`, I think we should extend the functionality we have in `RiscZeroCheats.sol` now, and provide this as a way to create a devnet programmatically using either `forge script` with a script provided in this repo, or importing and using that that deploy function in a  larger script, as we might be in the Boundless repo.

https://github.com/risc0/risc0-ethereum/blob/6a4eff5d906227f0911e1dbfb740645c6616cd4b/contracts/src/test/RiscZeroCheats.sol#L77-L89

PR #319, opened 2024-11-04
WEB3-213: add `SteelVerifier` to verify Steel commitments in the guest

Wollac:
This PR introduces the `SteelVerifier`, which acts as a built-in Steel `Contract` to verify Steel commitments. It is used like any other `Contract`, during the preflight step and in the guest. This functionality is currently marked unstable and must be enabled using the `unstable-verifier` feature.

The `TokenStats' example has been updated to use the verifier to sample the stats over two datapoints at different times.


nategraf:
Need to clean-up the way preflight is factored on the host

steel/src/block.rs
```diff
@@ -29,6 +29,20 @@ pub struct BlockInput<H> {
     ancestors: Vec<H>,
 }
 
+/// Implement [BlockHeaderCommit] for the unit type.
+/// This makes it possible to treat an `HostEvmEnv<D, H, ()>`, which is used for the [BlockInput]
+/// in the same way as any other `HostEvmEnv<D, H, BlockHeaderCommit>`.
+impl<H: EvmBlockHeader> BlockHeaderCommit<H> for () {
```
nategraf, line 35: You could use the following empty-type instead of unit-type. This is more clear and explicit, but I think implementing it on `()` also makes sense. Up to you.
```suggestion
struct DirectCommit;

/// Implement [BlockHeaderCommit] for the unit type.
/// This makes it possible to treat an `HostEvmEnv<D, H, DirectCommit>`, which is used for the [BlockInput]
/// in the same way as any other `HostEvmEnv<D, H, BlockHeaderCommit>`.
impl<H: EvmBlockHeader> BlockHeaderCommit<H> for DirectCommit {
```

steel/src/history/beacon_roots.rs
```diff
@@ -14,7 +14,24 @@
...
+pub const ADDRESS: Address = address!("000F3df6D732807Ef1319fB7B8bB8522d0Beac02");
+
+/// Hash of the contract's address, where the contract is deployed.
+const ADDRESS_HASH: B256 =
+    b256!("37d65eaa92c6bc4c13a5ec45527f0c18ea8932588728769ec7aecfe6d9f32e42");
+/// Hash of the deployed EVM bytecode.
+const CODE_HASH: B256 = b256!("f57acd40259872606d76197ef052f3d35588dadf919ee1f0e3cb9b62d3f4b02c");
 
 /// Enum representing possible errors that can occur within the `BeaconRootsContract`.
 #[derive(Debug, thiserror::Error)]
```
nategraf, line 37: We might add `non_exhaustive` to allow us to add more options in the future without breaking the API.
```suggestion
#[derive(Debug, thiserror::Error)]
#[non_exhaustive]
```

steel/src/history/beacon_roots.rs
```diff
@@ -170,16 +121,113 @@ impl BeaconRootsContract {
...
+        Ok(Some(AccountInfo {
+            balance: account.balance,
+            nonce: account.nonce,
+            code_hash: account.code_hash,
+            code: None,
+        }))
+    }
+
+    fn code_by_hash(&mut self, _code_hash: B256) -> Result<Bytecode, Self::Error> {
+        unreachable!()
```
nategraf, line 157: Semantically, this is closer to being unimplemented
```suggestion
        // should never be called.
        unimplemented!()
```

steel/src/history/beacon_roots.rs
```diff
@@ -170,16 +121,113 @@ impl BeaconRootsContract {
...
+        // only allow accessing the beacon roots contract's address
+        assert_eq!(address, ADDRESS);
+        Ok(self
+            .storage_trie
+            .get_rlp(keccak256(index.to_be_bytes::<32>()))?
+            .unwrap_or_default())
+    }
+
+    fn block_hash(&mut self, _number: u64) -> Result<B256, Self::Error> {
+        unreachable!()
```
nategraf, line 171:
```suggestion
        // should never be called.
        unimplemented!()
```

steel/src/host/mod.rs
```diff
@@ -161,6 +177,32 @@ impl<D, H: EvmBlockHeader, C> HostEvmEnv<D, H, C> {
...
+    pub async fn into_input(self) -> Result<EvmInput<H>> {
+        let input = BlockInput::from_proof_db(self.db.unwrap(), self.header).await?;
+
+        Ok(EvmInput::Block(input))
+    }
+}
+
+impl<D, H: EvmBlockHeader, C: Clone + BlockHeaderCommit<H>> HostEvmEnv<D, H, C> {
+    /// Returns the [Commitment] used to validate the environment.
+    pub fn commitment(&self) -> Commitment {
```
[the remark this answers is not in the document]
nategraf: Personally, I think this is alright. In most cases, I don't think devs will even notice.

steel/src/verifier.rs
```diff
@@ -0,0 +1,233 @@
...
+            self.db = Some(db);
+
+            result
+        }
+    }
+}
+
+fn validate_block_number(header: &impl EvmBlockHeader, block_number: U256) -> anyhow::Result<u64> {
+    let block_number: u64 = block_number.saturating_to();
+    let diff = header.number().saturating_sub(block_number);
```
nategraf, line 146: I feel like this should be `try_into` and `checked_sub`. I suppose these cases would be caught be the subsequent checks (e.g. there is no block hash in the DB for a future block number) but it seems prudent to exist early here IMO.
```suggestion
    let block_number: u64 = block_number.try_into().context("invalid block number >= 2^64")?;
    let diff = header.number().checked_sub(block_number).context("block number is greater than header block number")?;
```

nategraf (approved):

steel/src/verifier.rs
Wollac: Addressed in 18070b4b707a57a93ed2c1529161fbc9defaa4fc
I kept the `saturating_sub` as underflows are caught be the `ensure` in the following line and IMHO produce the better error message.

steel/src/block.rs
Wollac: This would technically be a breaking change, and testing it led to quite a few required changes in op-steel. So I kept the uint type.

PR #327, opened 2024-11-07
Add SteelVerifier example

Wollac:
Extend the token-stats example to use two sample points with only a single commitment.


examples/token-stats/methods/guest/src/main.rs
```diff
@@ -26,18 +26,34 @@ fn main() {
...
+        .call_builder(&CometMainInterface::getSupplyRateCall { utilization })
+        .call()
+        ._0;
+
+    // Prepare the second `EvmEnv` for execution.
+    let input: EthEvmInput = env::read();
+    let env2 = input.into_env().with_chain_spec(&ETH_MAINNET_CHAIN_SPEC);
+
+    // Verify the first commitment.
+    SteelVerifier::new(&env2).verify(env1.commitment());
```
nategraf, line 48: It's not clear here how `env1` and `env2` and are related. I guess I'd assume `env2` needs to be a more recent block than `env1`, but the API does not make this obvious (I'm actually not even sure myself that this is correct).

nategraf (commented):
I really like this update to the example!

PR #328, opened 2024-11-12
WEB3-229: fix: run history corruption tests only when the feature is enabled

Wollac:
Otherwise the integration tests will fail to compile if the `unstable-history` feature is not enabled.


tzerrell:
Thank you!

nategraf (approved):

PR #333, opened 2024-11-20
Update Groth16Verifier.sol: fix snarkjs public signals check bug

andrewcoder666:
Fix snarkjs public signals check bug.
see https://github.com/iden3/snarkjs/pull/518


nategraf (changes requested):
Can you provide more detail showing this is indeed a bug, and not a required check. If I understand correctly, the impact here of this extra check is extra gas usage. The impact of removing the check when it's required is critical soundness failure.

andrewcoder666:
> Can you provide more detail showing this is indeed a bug, and not a required check. If I understand correctly, the impact here of this extra check is extra gas usage. The impact of removing the check when it's required is critical soundness failure.

hi @nategraf, according to the [risc0 stark_verify.circom file](https://github.com/risc0/risc0/blob/main/groth16_proof/groth16/stark_verify.circom), there are only 5 output signals ( public witness ) passed to the Groth16 verifier, therefore only first 5 elements should be checked before ecPairing check. The line deleted is meant to check the non-existing 6th element of the array parameter, which obviously is to access to the uninitialized memory since the array itself only contains 5 elements as shown in the [code](https://github.com/risc0/risc0-ethereum/blob/main/contracts/src/groth16/RiscZeroGroth16Verifier.sol#L168).

nategraf:
Looks like this is almost certainly correct. Thanks @andrewcoder666 for bringing this! I'm going to do a bit more due diligence before merging this.

capossele:
Related to https://github.com/risc0/risc0/pull/2708#pullrequestreview-2553129660

nategraf (approved):

PR #338, opened 2024-11-25
Merkle Path Cleanup

Wollac:
This PR proposes a code cleanup to the Merkle path calculation in #336.
Also, test_consistency took a few minutes to run, so the maximum size has been removed.


nategraf (approved):

nategraf:
Thanks

PR #339, opened 2024-11-26
Release 1.2.0-rc.1

nategraf:
This release builds off of the risc0-zkvm 1.2.0 pre-release https://github.com/risc0/risc0/releases/tag/v1.2.0-rc.1, and prepares for the 1.2 release of risc0-ethereum.


contracts/src/groth16/RiscZeroGroth16Verifier.sol
```diff
@@ -55,7 +55,7 @@ contract RiscZeroGroth16Verifier is IRiscZeroVerifier, IRiscZeroSelectable, Grot
     using SafeCast for uint256;
 
     /// Semantic version of the the RISC Zero system of which this contract is part.
-    string public constant VERSION = "1.2.0-alpha.1";
+    string public constant VERSION = "1.2.0";
```
Wollac, line 58: I am not sure whether that should be 1.2.0 or 1.2.0-rc.1?

Wollac (approved):

contracts/src/groth16/RiscZeroGroth16Verifier.sol
nategraf: I debated this, and went with the other direction for the previous release. However, I decided to go with the final version number here because we don't actually expect this contract to change again (it's just dependent on the control IDs), and we will likely deploy it before 1.2 is out of rc. Right now our public deployments reference `1.1.0-rc.3`, which is a bit odd
https://sepolia.etherscan.io/address/0xd9b0d07CeCd808a8172F21fA7C97992168f045CA#code#F1#L57

PR #340, opened 2024-11-26
WEB3-241: Bump development version to 1.3.0-alpha.1

nategraf:
- **bump versions and update release instructions**
- **update steel changelog**
- **add note about updating the CI branch**


.github/scripts/cargo-update.sh
Wollac, line 1: I agree that such a script makes a lot of sense, but should it be part of this PR?
And if so, .github is probably not the best folder since it is not run by the CI.

Wollac (approved):

PR #364, opened 2024-12-12
WEB3-261: Index root in VerifiedRoot

capossele:
It seems that only the parameters marked with indexed become topics that we can filter on. Non-indexed parameters are encoded as part of the data payload of the log entry and cannot be used for filtering.
It also adds the seal as part of the event as unfortunately, in cases where the event gets triggered by an internal transaction, for instance when using `submitRootAndFulfillBatch` the call data of the method is not available from the original transaction


nategraf (approved):
Thanks! Definitely an oversight from before

PR #365, opened 2024-12-12
WEB3-262: Add Selector, Receipt and Seal utilities

capossele:
Groth16 seal utility functions to easily convert an onchain seal into a risc0 Receipt


contracts/src/groth16.rs
```diff
@@ -12,9 +12,82 @@
...
+            return Err(anyhow::anyhow!(
+                "Invalid selector {} for fake receipt",
+                hex::encode(selector)
+            ));
+        };
+        Receipt::new(
+            InnerReceipt::Fake(FakeReceipt::new(claim)),
+            journal.as_ref().to_vec(),
+        )
+    } else {
```
nategraf, line 85: I don't think this function should handle fake receipts. It's currently an odd compromise, where it handles Groth16 and Fake receipts, but not set inclusion receipts and it would not handle e.g. Plonk receipts if we add them later. We probably needs two types of decoding. One would be a method that decodes to a specific type (e.g. Groth16, Fake, or SetInclusion), and another that decodes any seal in a list of known types. For the later, we will need a table included in the library mapping selectors to verifier parameter (digests). At the very least, that table should include Groth16, SetInclusion and Fake verifier parameters for the current version. We could additionally include previous versions in the this table as well. We've discussed this elsewhere, and it seems likely to be useful.

contracts/src/groth16/ControlID.sol
```diff
@@ -20,7 +20,7 @@
 pragma solidity ^0.8.9;
 
 library ControlID {
-    bytes32 public constant CONTROL_ROOT = hex"8cdad9242664be3112aba377c5425a4df735eb1c6966472b561d2855932c0469";
+    bytes32 public constant CONTROL_ROOT = hex"ffe166017b1ab460995b45510524c60e3756344feff2414f44cda25c33a78058";
```
nategraf, line 23: Let's make control root changes in a separate PR if needed.

contracts/src/receipt.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use crate::{
+    encode_seal,
+    groth16::decode_groth16_seal,
+    selector::{SelecorType, Selector, SelectorError},
+};
+
+/// Wrapper around different types of receipts.
+pub enum ReceiptType {
+    Risc0(Risc0Receipt),
```
nategraf, line 31:
```suggestion
/// Extension of the base [risc0-zkvm::Receipt] type.
pub enum Receipt {
    Base(risc0_zkvm::Receipt),
```

contracts/src/receipt.rs
```diff
@@ -0,0 +1,109 @@
...
+    SetInclusion(SetInclusionReceipt<ReceiptClaim>),
+}
+
+impl ReceiptType {
+    /// Encode the receipt as a seal.
+    pub fn encode_seal(&self) -> Result<Vec<u8>, SetInclusionEncodingError> {
```
nategraf, line 37: To be a bit more specific. Folks might be looking for ABI as a keyword as well.
```suggestion
    pub fn abi_encode_seal(&self) -> Result<Vec<u8>, SetInclusionEncodingError> {
```

contracts/src/receipt.rs
```diff
@@ -0,0 +1,109 @@
...
+        match self {
+            ReceiptType::Risc0(receipt) => {
+                encode_seal(receipt).map_err(|_| SetInclusionEncodingError::UnsupportedReceiptType)
+            }
+            ReceiptType::SetInclusion(receipt) => receipt.abi_encode_seal(),
+        }
+    }
+
+    /// Get the receipt if it is a RISC0 receipt.
```
nategraf, line 46:
```suggestion
    /// Get the receipt if it is a base [risc0_zkvm::Receipt].
```

contracts/src/receipt.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub fn receipt(&self) -> Option<Risc0Receipt> {
+        match self {
+            ReceiptType::Risc0(receipt) => Some(receipt.clone()),
```
nategraf, line 49: Is there a reason this would not work? With this, they could clone it if needed or use it as a reference.
```suggestion
    pub fn receipt(&self) -> Option<&risc0_zkvm::Receipt> {
        match self {
            ReceiptType::Risc0(receipt) => Some(receipt),
```

contracts/src/selector.rs
```diff
@@ -0,0 +1,92 @@
...
+use hex::FromHex;
+use risc0_zkvm::sha::Digest;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+#[non_exhaustive]
+pub enum SelectorError {
+    #[error("Unsupported selector")]
+    UnsupportedSelector,
+    #[error("Selector (0) does not have verifier parameters")]
```
nategraf, line 24: I think this needs to be curly braces. I think.
```suggestion
    #[error("Selector {0} does not have verifier parameters")]
```

contracts/src/selector.rs
```diff
@@ -0,0 +1,92 @@
...
+    NoVerifierParameters(Selector),
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum SelecorType {
```
nategraf, line 30:
```suggestion
pub enum SelectorType {
```

contracts/src/receipt.rs
```diff
@@ -0,0 +1,109 @@
...
+    #[error("Selector error: {0}")]
+    SelectorError(#[from] SelectorError),
+    #[error("Decoding error: {0}")]
+    SetInclusionError(#[from] SetInclusionDecodingError),
+    #[error("Decoding error: {0}")]
+    Anyhow(#[from] anyhow::Error),
+}
+
+/// Decode a seal into a receipt.
+pub fn decode_seal(
```
nategraf, line 80: I'd recommend we split this into two function APIs. We can have an `fn decode_seal(seal: Bytes, image_id: impl Into<Digest>, journal: impl Into<Vec<u8>>)` and an `fn decode_seal_with_claim(seal: Bytes, claim: ReceiptClaim, journal: impl Into<Vec<u8>>)`. Reason being, most users don't ever use `ReceiptClaim` directly or really know much about it. Given an image ID and journal, we can call `ReceiptClaim::ok` to get the receipt claim, assuming the execution ended normally (e.g. did not panic) which covers every use case so far.

contracts/src/selector.rs
```diff
@@ -0,0 +1,92 @@
...
+            0x00000000 => Ok(Selector::FakeReceipt),
+            0x50bd1769 => Ok(Selector::Groth16V1_1),
+            0xc101b42b => Ok(Selector::Groth16V1_2),
+            0xbfca9ccb => Ok(Selector::SetVerifierV0_1),
+            _ => Err(SelectorError::UnsupportedSelector),
+        }
+    }
+}
+
+impl Selector {
```
nategraf, line 60: Let's add a note to `RELEASE.md` to update this when the control root changes.

contracts/src/selector.rs
```diff
@@ -0,0 +1,92 @@
...
+                "50bd1769093e74abda3711c315d84d78e3e282173f6304a33272d92abb590ef5",
+            )
+            .unwrap()),
+            Selector::Groth16V1_2 => Ok(Digest::from_hex(
+                "c101b42bcacd62e35222b1207223250814d05dd41d41f8cadc1f16f86707ae15",
+            )
+            .unwrap()),
+            Selector::SetVerifierV0_1 => Ok(Digest::from_hex(
+                "bfca9ccb59eb38b8c78ddc399a734d8e0e84e8028b7d616fa54fe707a1ff1b3b",
+            )
```
nategraf, line 76: Would you go ahead and add the v0.2 digest?

contracts/src/set_verifier.rs
```diff
@@ -116,4 +133,110 @@ where
...
+
+    async fn get_latest_block(&self) -> Result<u64> {
+        self.instance
+            .provider()
+            .get_block_number()
+            .await
+            .context("Failed to get latest block number")
+    }
+
+    /// Query the VerifiedRoot event based on the root and block options.
```
nategraf, line 150:
```suggestion
    /// Query the VerifiedRoot event based on the root and block options.
    ///
```

contracts/src/set_verifier.rs
```diff
@@ -116,4 +133,110 @@ where
...
+        let start_block = lower_bound.unwrap_or(upper_block.saturating_sub(
+            self.event_query_config.block_range * self.event_query_config.max_iterations,
+        ));
+
+        // Loop to progressively search through blocks
+        for _ in 0..self.event_query_config.max_iterations {
+            // If the current end block is less than or equal to the starting block, stop searching
+            if upper_block <= start_block {
+                break;
+            }
```
nategraf, line 173: We should search the single block if they are equal, no? Or is upper-bound not inclusive. If it is not, then it seems like the default above should be latest + 1
```suggestion
            // If the current end block is less than to the starting block, stop searching
            if upper_block < start_block {
                break;
            }
```

contracts/src/set_verifier.rs
```diff
@@ -116,4 +133,110 @@ where
...
+    /// The default range is set to 100 blocks for each iteration, and the default maximum number of
+    /// iterations is 100. This means that the search will cover a maximum of 10,000 blocks.
+    /// Optionally, you can specify a lower and upper bound to limit the search range.
+    async fn query_verified_root_event(
+        &self,
+        root: B256,
+        lower_bound: Option<u64>,
+        upper_bound: Option<u64>,
+    ) -> Result<Bytes> {
+        let mut upper_block = upper_bound.unwrap_or(self.get_latest_block().await?);
```
nategraf, line 163: `unwrap_or` is evaluated eagerly, meaning this will make an API call every time, even if an upper-bound is provided. I wish there was a native async version of `unwrap_or_else`, but there is not.
```suggestion
        let mut upper_block = if let Some(upper_bound) = upper_bound else { self.get_latest_block().await? };
```

contracts/src/set_verifier.rs
```diff
@@ -116,4 +133,110 @@ where
...
+            .context("Failed to get latest block number")
+    }
+
+    /// Query the VerifiedRoot event based on the root and block options.
+    /// For each iteration, we query a range of blocks.
+    /// If the event is not found, we move the range down and repeat until we find the event.
+    /// If the event is not found after the configured max iterations, we return an error.
+    /// The default range is set to 100 blocks for each iteration, and the default maximum number of
+    /// iterations is 100. This means that the search will cover a maximum of 10,000 blocks.
+    /// Optionally, you can specify a lower and upper bound to limit the search range.
```
nategraf, line 156:
```suggestion
    /// The default range is set to 100 blocks for each iteration, and the default maximum number of
    /// iterations is 1000. This means that the search will cover a maximum of 100,000 blocks (~14 days at 12s block times).
    /// Optionally, you can specify a lower and upper bound to limit the search range.
```

contracts/src/set_verifier.rs
```diff
@@ -116,4 +133,110 @@ where
 
         Ok((image_id, image_url))
     }
+
+    /// Returns the seal if of the given verified root.
+    pub async fn get_verified_root_seal(&self, root: B256) -> Result<Bytes> {
```
nategraf, line 138: I like the word `query` here to indicate a network call will occur. I think `fetch` is also good.
```suggestion
    pub async fn query_verified_root_seal(&self, root: B256) -> Result<Bytes> {
```

contracts/src/set_verifier.rs
```diff
@@ -116,4 +133,110 @@ where
...
+            // Move the upper_block down for the next iteration
+            upper_block = lower_block.saturating_sub(1);
+        }
+
+        // Return error if no logs are found after all iterations
+        bail!("VerifiedRoot event not found for root {:?}", root);
+    }
+
+    /// Decodes a seal into a [SetInclusionReceipt] including a [risc0_zkvm::Groth16Receipt] as its root.
+    pub async fn get_receipt(
```
nategraf, line 203: Same comment about `query` or `fetch`. Also similar comment about providing a based option that takes the (image_id, journal) and an "advanced" option that takes the receipt claim.
```suggestion
    pub async fn query_receipt(
```

nategraf (commented):
I like this a lot! Modulo my comments, I think we should merge this as is. I did mention there is some opportunity to consolidate with the upcoming work on composition, and I'll take care or that when I get back to composition next week. I think we should not wait to merge this.

contracts/src/set_verifier.rs
capossele: went with `fetch`

nategraf (approved):
Nice!

PR #370, opened 2024-12-17
WEB3-294: Update deployment.toml to mark 1.2 mainnet verifiers as routable

nategraf:
Timelock controller execute transactions have now cleared, and so the 1.2 verifiers are live on mainnets.


nategraf:
Converting to draft since I added from changes to CI that needs to be considered.

capossele (approved):

PR #381, opened 2025-01-09
BM-85: Refactor aggregation set builder for incremental tree building

nategraf:
The aggregation set builder is used extensively in Boundless and one of the main bottlenecks right now with service large numbers of small requests is that aggregation is done in a series of sequential steps.
In theory, these steps could be parallelized but this is complicated in practice.

This PR refactors the set-builder to be able to accept multiple claims per execution, and can aggregated a given vector of receipts in one-step if immediate finalization is desired.
It does so by using a Merkle mountain range structure to encode intermediate states in building the final tree.
When combined with the work on parallelized assumption resolution (through the `union` program) this should take advantage of existing parallelized workflows in the prover cluster to achieve much better e2e times on receipt aggregation, and therefore on request fulfillment.

Related: BM-85


nategraf:
PR is in draft as the test coverage is not yet sufficient.

build/src/lib.rs
```diff
@@ -108,7 +108,8 @@ pub fn generate_image_id_sol(guests: &[GuestListEntry]) -> Result<Vec<u8>> {
         .iter()
         .map(|guest| {
             let name = guest.name.to_uppercase().replace('-', "_");
-            let image_id = hex::encode(guest.image_id);
+            // NOTE: Works with both risc0-build v1 and v2
+            let image_id = hex::encode(bytemuck::cast::<_, [u8; 32]>(guest.image_id));
```
nategraf, line 112: This is an issue when using the latest `risc0-build` on `main` which bumps to v2 and changes the `image_id` field from `[u32; 8]` to `Digest`. This line is chosen to work with both versions.
https://github.com/risc0/risc0/commit/881e512732eca72849b2d0e263a1242aba3158af#diff-4bac0a581b7140eead855e0ed6debc8734ce2b49e3d5317f53fb13c5aa1335e3L169-R181

aggregation/src/lib.rs
```diff
@@ -52,45 +53,309 @@ alloy_sol_types::sol! {
...
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(test, derive(PartialEq, Eq))]
+struct Peak {
+    /// Digest for the root of the Merkle subtree committed to by this peak.
+    digest: Digest,
+    /// An upper-bound on the depth of the subtree rooted rooted at this peak.
+    ///
+    /// It is expressed as the total height of the subtree - 1, such that a peak with a single node
+    /// (i.e. a leaf) has a max_depth value of 0.
+    max_depth: u8,
```
Wollac, line 162: We should mention 0xff for a finalized MMR here

aggregation/src/lib.rs
```diff
@@ -52,45 +53,309 @@ alloy_sol_types::sol! {
...
+            Some(peak) if peak.max_depth > new_peak.max_depth => {
+                self.0.push(new_peak);
+            }
+            // If the tail subtree is the same size, combine them and recurse.
+            Some(peak) if peak.max_depth == new_peak.max_depth => {
+                // Will never panic, since we've already checked that there is at least one peak.
+                let peak = self.0.pop().unwrap();
+                self.push_peak(Peak {
+                    digest: commutative_keccak256(&peak.digest, &new_peak.digest),
+                    max_depth: peak.max_depth + 1, //TODO this is fallible
```
Wollac, line 230: Even though 2^254 is very, very large, we could just add an assert(peak.max_depth < 0xfe) here for clarity.

aggregation/src/lib.rs
```diff
@@ -52,45 +53,309 @@ alloy_sol_types::sol! {
...
+///
+/// Each entry in the list is a pair of (digest, max-depth), where max-depth tracks an upper bound
+/// on the size of the subtree for which the digest is the root. The largest subtree is at index 0,
+/// the smallest at index len - 1.
+///
+/// Note that the max size of the internal vec of digests (peaks) is equal to log_2 n where n is
+/// the number of leaves in the tree.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(test, derive(PartialEq, Eq))]
+pub struct MerkleMountainRange(Vec<Peak>);
```
Wollac, line 151: In other implementations of MMRs in the R0 code base, we usually use Vec<Option<Digest>> to represent it, where the i-th digest corresponds to the Merkle trie with height i in the range, if it exists.
Both representation have pros and cons, just wanted to mention it.

aggregation/src/lib.rs
```diff
@@ -52,45 +53,309 @@ alloy_sol_types::sol! {
...
+    pub fn is_finalized(&self) -> bool {
+        // If the peak has a max-depth of 255, then the mountain range is finalized and no new
+        // peaks can be pushed to it. Note that this state can only be achieved by calling
+        // `finalize` since it is computationally infeasible to push 2^256 nodes onto the MMR,
+        // although it is theoretically consistent that if an MMR reached a state of being a
+        // single peak with max depth value of 255, it would be naturally finalized.
+        self.0
+            .first()
+            .map(|peak| peak.max_depth == u8::MAX)
+            .unwrap_or(false)
```
Wollac, line 284:
```suggestion
            .map_or(false, |peak| peak.max_depth == u8::MAX)
```

Shouldn't clippy complain about this?

aggregation/src/lib.rs
```diff
@@ -52,45 +53,309 @@ alloy_sol_types::sol! {
...
-            } => commutative_keccak256(left_set_root, right_set_root),
+    /// Decode the specialized journal encoding. See [MerkleMountainRange::encode].
+    pub fn decode(bytes: impl AsRef<[u8]>) -> Result<Self, DecodingError> {
+        // Read the first 32 bytes as the bitmap.
+        let (mut chunk, mut bytes) = bytes
+            .as_ref()
+            .split_at_checked(U256::BYTES)
+            .ok_or(DecodingError::UnexpectedEnd)?;
+        let bitmap = U256::from_le_slice(chunk);
+        if bitmap > (U256::from(1) << u8::MAX) {
```
Wollac, line 316: Does the Compiler understands that this is a constant or will it be runtime evaluated?
I would go with `uint!(1_U256 << u8::MAX)` for added clarity.

Wollac (approved):

aggregation/src/lib.rs
```diff
@@ -52,45 +53,309 @@ alloy_sol_types::sol! {
...
+    /// Will return an error if the [MerkleMountainRange] on the [GuestState] is already
+    /// finalized, as no more claims may be added and the guest would reject this input.
+    pub fn into_input(
+        self,
+        claims: Vec<ReceiptClaim>,
+        finalize: bool,
+    ) -> Result<GuestInput, Error> {
+        if self.mmr.is_finalized() {
+            return Err(Error::FinalizedError);
+        }
```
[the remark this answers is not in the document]
nategraf: Since this is not the standard `Into` trait anyway I lean towards returning the error here. It's easy enough for the host to `unwrap` if the believe they'll never hit that case.

aggregation/src/lib.rs
nategraf: Indeed, ya! I wrote it that way at first, but liked the way it turned out this way instead.

aggregation/src/lib.rs
nategraf: 6c34c00

aggregation/src/lib.rs
nategraf: Used `checked_add` here. Thanks for catching this, I meant to address it.
928fccc

aggregation/src/lib.rs
nategraf: Nice! I didn't know about that macro.
6627c08

PR #383, opened 2025-01-10
WEB3-281: update create-steel-app to 1.2

sashaaldrick:
* updated create-steel-app to use 1.2 by default
* also added a line to the README to make sure that Foudry is updated to avoid a relatively old bug (most end users should be fine, but I ran into this issue)


Wollac (approved):

capossele (approved):

capossele:
@sashaaldrick I added some additional changes to make CI happy

PR #386, opened 2025-01-14
WEB3-285: Sasha/add steel history docs

sashaaldrick:
* added `steel/docs/steel-history.md` for documentation about Steel history feature
* added links to `steel/README.md` for navigation

very happy to consolidate sections + structure and keen to hear feedback


steel/docs/steel-history.md
```diff
@@ -0,0 +1,94 @@
...
+
+With the Steel history feature enabled, the developer must specify both the execution block *and* the commitment block:
+
+```rust
+/// WITH STEEL HISTORY
+let builder = EthEvmEnv::builder()
+    .provider(provider.clone())
+    .block_number_or_tag(args.execution_block);
+
+#[cfg(any(feature = "beacon", feature = "history"))]
```
Wollac, line 35: The history and the beacon feature were just added to the example so that the same example can be used to test all these different variants.
Do you think it is best to leave those cfg's in to match the example 100% or just remove the cfg lines to make the code snippet cleaner?

steel/docs/steel-history.md
```diff
@@ -0,0 +1,94 @@
...
+In terms of compute, the number of cycles for the Steel guest will also increase linearly with the number of blocks between the commitment block and the execution block. For every extra 24 hours of history, this is around 1 million cycles in the Steel guest.
+
+## Enabling Steel History
+
+To enable Steel History, you can add these feature flags to the relevant `Cargo.toml` file:
+
+```toml
+[features]
+history = ["risc0-steel/unstable-history"]
+beacon = []
```
Wollac, line 78: Technically it is only required to enable the unstable-history feature in Steel how this is done in the application and wether it is called history is not relevant.

steel/docs/steel-history.md
```diff
@@ -0,0 +1,94 @@
...
+let mut contract = Contract::preflight(args.token_contract, &mut env);
+```
+
+The execution block is the block from which the view call state is retrieved. The commitment block is the block used for the Steel commitment. The commitment block has to fall within the 24 hour time window necessary for Steel commitment validation on-chain, but *crucially* the execution block can go further back on the scale of days, weeks or even months. 
+
+## How does Steel history work?
+
+The execution and commitment blocks are fundamentally related; the execution block should always be an ancestor of the commitment block. Therefore, it is possible to prove that the commitment block block root is canonical with the execution block block root by validating each beacon block root in between the two blocks.
+
+Steel handles this verification automatically when the history feature is enabled. Steel history works backwards from the commitment block to the execution block with consecutive calls to the beacon roots contract; validating a beacon root is a single call for every 24 hours of history and this step takes approximately 1M cycles within the Steel guest.
```
Wollac, line 52: Should we add the actual feature name `unstable-history` here?

steel/docs/steel-history.md
```diff
@@ -0,0 +1,94 @@
...
+## How does Steel history work?
+
+The execution and commitment blocks are fundamentally related; the execution block should always be an ancestor of the commitment block. Therefore, it is possible to prove that the commitment block block root is canonical with the execution block block root by validating each beacon block root in between the two blocks.
+
+Steel handles this verification automatically when the history feature is enabled. Steel history works backwards from the commitment block to the execution block with consecutive calls to the beacon roots contract; validating a beacon root is a single call for every 24 hours of history and this step takes approximately 1M cycles within the Steel guest. 
+
+Ultimately, Steel will check the integrity of the view call data in the execution block by:
+
+1. proving that the execution block is a canonical ancestor of the commitment block
+2. carrying out the standard Steel commitment validation on-chain proving the integrity of the block root for the commitment block
```
Wollac, line 57: Very strictly speaking, this step of verifying the standard Steel commitment is not part of Steel as it happens on-chain

steel/docs/steel-history.md
```diff
@@ -0,0 +1,94 @@
...
+
+There is a hard limit on how far back in time you can place the execution block: the entire validation procedure depends on [EIP-4788] which was introduced with the [Cancun upgrade] on March 13 2024; this is the furthest that Steel history can go back.
+
+## How much does Steel history cost? 
+
+For the host, the developer needs to specify valid RPC URLs for both an archive execution node and a beacon node. You can see an example in the [publisher.rs] CLI args for the [erc20-counter] example.
+
+The greatest API cost will likely be from the beacon API endpoint. For each block between the commitment block and the execution block, Steel will query the full beacon block for verification. Please bear in mind that the wider the gap between the execution block and the commitment block, the larger the load on the beacon endpoint will be. 
+
+In terms of compute, the number of cycles for the Steel guest will also increase linearly with the number of blocks between the commitment block and the execution block. For every extra 24 hours of history, this is around 1 million cycles in the Steel guest.
```
Wollac, line 69: The cycle cost was before accelerators, when the keccak accelerator has landed it should significantly less expensive.

Wollac (approved):

steel/docs/steel-history.md
nategraf: I'd definitely suggest we remove the `cfgs` from the example. In a user application, we expect they will select an approach and chose one or these other, rather than having feature flags like this.

nategraf (approved):
Looks good! Thanks for writing up a good page on this! I did a pass in my editor and created a PR with suggested edits.
https://github.com/risc0/risc0-ethereum/pull/388

steel/docs/steel-history.md
sashaaldrick: Interesting, should we update the publisher app? Is this the recommended way of enabling history or no? I understand you can also compile with feature flags, but I think that would be a less used method (though I could be wrong).

steel/docs/steel-history.md
sashaaldrick: That's a good point, I have fixed this to reflect that change with:

> Ultimately, Steel will check the integrity of the view call data in the execution block by proving that the execution block is a canonical ancestor of the commitment block. 
Once on-chain, succesfully validating the Steel commitment will prove the integrity of the block root for the commitment block.

steel/docs/steel-history.md
sashaaldrick: thanks, added a note on that!

steel/docs/steel-history.md
sashaaldrick: Ah, @nategraf sorted this out in his PR, I understand what you meant now @Wollac. Thanks and resolved

steel/docs/steel-history.md
sashaaldrick: I understand the reasoning for removing `cfgs`, and will keep that in mind for the future. Thanks both

PR #394, opened 2025-01-17
Bump version for release v1.3.0

capossele:


nategraf (approved):
Looks like there is an orphaned lock file. I removed it in https://github.com/risc0/risc0-ethereum/pull/394/commits/dda3a2afa561e2e7d62bef058c76612dbde43337

PR #405, opened 2025-01-21
WEB3-307: feat: Add event support to Steel

Wollac:
Adds support to proof the inclusion of events in Steel.


Wollac:
Replaced by #409

PR #409, opened 2025-01-23
WEB3-315: feat: Add EVM Event support to Steel

Wollac:
### Description

This PR introduces the capability to query Ethereum events within Steel. It allows developers to efficiently retrieve specific event logs emitted by smart contracts and confidently use them in the guest.
Its usage is very similar to the existing `Contract` which is used to execute contract calls.

**Key Features:**

*   **Event Querying:**  Enables querying for events based on their type (e.g., `IERC20::Transfer`).
*   **ZK Verification:**  Guarantees the integrity and correctness of event query results within the guest environment. To achieve this, the input includes all relevant receipts, unless the header's Bloom filter indicates that no matching logs are present.
*   **Guest-Side Verification:** The guest can verify the query results against the header's Bloom filter field or the Receipts root field, ensuring trustworthiness.

### Performance

Event queries on Ethereum Mainnet, involving approximately 100-300 receipts per block, are executed with an average of 4 million cycles utilizing the Keccak accelerator.

### Usage

The following code snippet illustrates how to query for ERC20 `Transfer` events using the new feature:

**Host-Side (Fetching EVM Input):**

```rust
use alloy_primitives::address;
use alloy_sol_types::sol;
use steel_evm::{EthEvmEnv, Event};

let contract_address = address!("dAC17F958D2ee523a2206206994597C13D831ec7"); // USDT contract address
sol! {
    interface IERC20 {
        event Transfer(address indexed from, address indexed to, uint256 value);
    }
}

// Host:
let url = "https://ethereum-rpc.publicnode.com".parse()?;
let mut env = EthEvmEnv::builder().rpc(url).build().await?;
let event = Event::preflight::<IERC20::Transfer>(&mut env).address(contract_address);
event.query().await?;

let evm_input = env.into_input().await?;
```

**Guest-Side (Querying and Verifying Events):**

```rust
use alloy_primitives::address;
use alloy_sol_types::sol;
use steel_evm::{EthEvmEnv, Event};

// Assuming `evm_input` is passed to the guest
let env = evm_input.into_env();
let event = Event::new::<IERC20::Transfer>(&env).address(contract_address); // USDT contract address
let logs = event.query();

// ... further processing of the retrieved `logs` within the guest ...
```

### Example

A practical example is included, demonstrating how to calculate the total amount of USDT transferred in a specific block by querying and aggregating the `Transfer` events emitted by the USDT ERC20 contract. This showcases the real-world applicability of event querying with ZK verification.


crates/steel/src/lib.rs
```diff
@@ -203,6 +229,12 @@ pub trait EvmBlockHeader: Sealable {
     fn timestamp(&self) -> u64;
     /// Returns the state root hash.
     fn state_root(&self) -> &B256;
+    #[cfg(feature = "unstable-event")]
```
Wollac, line 232: This adds new fields we require from a block. This is a breaking change of the EvmBlockHeader trait.
And, unfortunately, it is difficult to hide this change behind feature flags.

crates/steel/src/host/db/proof.rs
```diff
@@ -311,3 +366,53 @@ fn add_proof(
...
+            // Unfortunately ReceiptResponse does not implement ReceiptEnvelope, so we have to
+            // manually convert it. We convert to a TransactionReceipt which is the default and
+            // works for Ethereum-compatible networks.
+            // Use serde here for the conversion as it is much safer than mem::transmute.
+            // TODO(https://github.com/alloy-rs/alloy/issues/854): use ReceiptEnvelope directly
+            let json = serde_json::to_value(rpc_receipt).context("failed to serialize")?;
+            let tx_receipt: TransactionReceipt = serde_json::from_value(json)
+                .context("failed to parse as Ethereum transaction receipt")?;
+
+            Ok(tx_receipt_to_envelope(tx_receipt))
```
Wollac, line 387: It is very unfortunate that in the current version of Alloy a ReceiptResponse does not implement the ReceiptEnvelop (as the TransactionResponse does for example).
So we have to convert it manually. I tried two approaches:
- Switch from the general `Network` to `Ethereum`, where the trait boundaries are correct. However, this completely changes the bounds in Steel and breaks op-steel support.
- Cast" the receipt response. The current code does this in a memorysafe way with serde-json. This is ugly, but at least it works for everything that is Ethereum compatible. (It will not otherwise cause any weird bugs in the guest, as the resulting receipt root is verified.)

Eventually, this can only be fixed when https://github.com/alloy-rs/alloy/issues/854 is fixed upstream.

crates/steel/src/event.rs
```diff
@@ -0,0 +1,193 @@
...
+impl<H: EvmBlockHeader> Event<(), &GuestEvmEnv<H>> {
+    /// Constructor for executing an event query for a specific Solidity event.
+    pub fn new<S: SolEvent>(env: &GuestEvmEnv<H>) -> Event<S, &GuestEvmEnv<H>> {
+        Event {
+            filter: event_filter::<S, H>(env.header()),
+            env,
+            phantom: PhantomData,
+        }
+    }
+}
```
Wollac, line 82: Implementing this for `Event<(), E>` makes it possible to initialize with
```rust
Event::new::<IERC20::Transfer>(&env)
```

Instead of having to also specify `E`.

In general I decided to assign a fixed SolEvent to a `Steel`. This allows the corresponding Event::query() to return the actual events of the correct type.
An alternative would be to specify an arbitrary filter in the query step and leave all casting and conversion to the user.

crates/steel/src/event.rs
```diff
@@ -0,0 +1,193 @@
...
+            logs.iter()
+                .map(|log| Ok(S::decode_log(log, false)?))
+                .collect()
+        }
+    }
+}
+
+/// Creates an event filter for a specific Solidity event and block header.
+fn event_filter<S: SolEvent, H: EvmBlockHeader>(header: &Sealed<H>) -> Filter {
+    assert!(!S::ANONYMOUS, "Anonymous events not supported");
```
Wollac, line 189: It is not possible to cleanly handle anonymous events with this approach. Since do not have a signature success of the decode_log cannot be guaranteed.

crates/steel/src/block.rs
```diff
@@ -27,6 +28,7 @@ pub struct BlockInput<H> {
     storage_tries: Vec<MerkleTrie>,
     contracts: Vec<Bytes>,
     ancestors: Vec<H>,
+    receipts: Option<Vec<Eip2718Wrapper<ReceiptEnvelope>>>,
```
Wollac, line 31: The `ReceiptEnvelope` cannot be serialized using bincode or similar, so it must be RLP encoded. However, since the hash calculation also relies on the RLP encoding, this may actually be a performance improvement.

crates/steel/src/event.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// Sets the address to query with this filter.
+    ///
+    /// See [`Filter::address`].
+    pub fn address<A: Into<ValueOrArray<Address>>>(mut self, address: A) -> Self {
+        self.filter.address = address.into().into();
+        self
+    }
+
+    /// Sets the 1st indexed topic.
+    pub fn topic1<TO: Into<Topic>>(mut self, topic: TO) -> Self {
```
nategraf, line 110: Is is possible to additionally provide a more type-safe option using `SolEvent::TopicList`? It would be nice, but I can't tell from their docs how to make this work myself.

crates/steel/src/lib.rs
```diff
@@ -203,6 +229,12 @@ pub trait EvmBlockHeader: Sealable {
...
+    /// Returns the receipts root hash of the block.
+    fn receipts_root(&self) -> &B256;
+    #[cfg(feature = "unstable-event")]
+    /// Returns the logs bloom filter of the block
+    fn logs_bloom(&self) -> &Bloom;
```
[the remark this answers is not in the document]
nategraf: Maybe we should do that along with a 2.0 version bump. I am not opposed to bumping to 2.0.

crates/steel/src/lib.rs
nategraf: I think this is probably ok. I would be a bit surprised if anyone is implementing this themselves right now, and if anyone screams we can roll back the release and bump to 2.0 when we release this.

nategraf (approved):
Very nice! I think this is going to be very useful. We should send show to Mendi as soon as it's in a release.

PR #410, opened 2025-01-23
WEB3-316: Fix byte ordering in encoding of aggregation guest state

nategraf:


capossele (approved):
Lost my bit of time as well with similar bugs 🙈

PR #416, opened 2025-01-27
WEB3-282: feat: Allow querying blocks by hash

### LINKED ISSUE #384: [Feature] Allow querying blocks by hash
## Feature

Blocks can currently be queried by height or tag, where it would be nice to also have hash https://github.com/risc0/risc0-ethereum/blob/917180dd915a283e0a29c9df7c7970b6eca9ffc9/steel/src/host/mod.rs#L50-L65

## Motivation

Some users have a block hash, and doing the lookup/sanity check is a bit unideal

Wollac:
Fixes #384 and WEB3-282

### DIFF
--- a/crates/steel/src/host/builder.rs
+++ b/crates/steel/src/host/builder.rs
@@ -24,8 +24,8 @@ use crate::{
     },
     EvmBlockHeader, EvmEnv,
 };
-use alloy::eips::BlockId as AlloyBlockId;
 use alloy::{
+    eips::BlockId as AlloyBlockId,
     network::{
         primitives::{BlockTransactionsKind, HeaderResponse},
         BlockResponse, Ethereum, Network,
--- a/crates/steel/src/host/builder.rs
+++ b/crates/steel/src/host/builder.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use super::BlockId;
 use crate::{
     beacon::BeaconCommit,
     config::ChainSpec,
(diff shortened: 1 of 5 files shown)


crates/steel/src/host/builder.rs
Wollac, line 1: For the (unstable) history feature we only have the method [commitment_block(self, BlockNumberOrTag)](https://github.com/risc0/risc0-ethereum/blob/25b79c67e47b2db9a8fadc9b5487d968f516cc50/crates/steel/src/host/builder.rs#L246).
Should we make this consistent with `block_number`, `block_number_or_tag`, `block_hash` for the execution block? Or would that complicate the API unnecessarily (related to WEB3-272)

crates/steel/src/host/mod.rs
```diff
@@ -28,22 +23,80 @@ use crate::{
...
 pub mod db;
 
 pub use builder::EvmEnvBuilder;
 
+/// A Block Identifier.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+enum BlockId {
+    /// A block hash
+    Hash(BlockHash),
+    /// A block number
```
nategraf, line 54:
```suggestion
    /// A block number or tag (e.g. latest)
```

crates/steel/src/host/builder.rs
nategraf: It's a balance for sure, but if possible that would be beneficial. Block hashes may be preferred in some situations, since they are unambiguous (incl in contexts where multiple chains may be referred to, or in a context where reorgs must be handled). It's also not the end of the world to say they need to get the block number and pass that in instead 🤷

nategraf (approved):

crates/steel/src/host/builder.rs
Wollac: Added the missing methods in cca36dac81cf1a2be2a541d032444b1c2d14ce91

PR #419, opened 2025-01-27
WEB3-321: feat: deploy steel docs on github pages

nahoc:
This deploys a custom docs.rs like doc site for steel using GitHub pages and `cargo doc`

### DIFF
--- a/.github/workflows/steel-documentation.yml
+++ b/.github/workflows/steel-documentation.yml
@@ -63,7 +63,8 @@ jobs:
           for dir in examples docs images assets; do
             if [ -d ../../$dir ]; then
               ln -s ../../$dir $dir
-              # Also create in nested structure
+              # Also create in nested structure for deeper doc pages
+              mkdir -p risc0_steel
               ln -s ../../../$dir risc0_steel/$dir
             fi
           done
--- a/.github/workflows/steel-documentation.yml
+++ b/.github/workflows/steel-documentation.yml
@@ -3,9 +3,6 @@ name: Deploy Steel Documentation
 on:
   push:
     branches: ["main"]
-  pull_request:
-  # Allow manual trigger
-  workflow_dispatch:
 
 # Sets permissions of the GITHUB_TOKEN to allow deployment to GitHub Pages
 permissions:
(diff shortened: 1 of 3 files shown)


.github/workflows/steel-documentation.yml
```diff
@@ -0,0 +1,55 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      
+      - name: Setup Pages
+        uses: actions/configure-pages@v4
+      
+      - uses: risc0/risc0/.github/actions/rustup@main    
+
+      - name: Build documentation
+        run: cargo doc --no-deps --document-private-items
```
nategraf, line 34: I don't think we want to show private items here. They can read the source code if they'd like that level of detail.
```suggestion
        run: cargo doc --no-deps
```

nategraf (commented):
When we are building for publishing, we'd like to use `--cfg docsrs` and `--all-features` so that features under flags like `unstable-history` show up and are marked as requiring a feature flag.
https://github.com/risc0/risc0-ethereum/blob/main/.github/workflows/main.yml#L251

.github/workflows/steel-documentation.yml
```diff
@@ -0,0 +1,55 @@
+name: Deploy Steel Documentation
+
+on:
+  push:
+    branches: ["main"]
```
nategraf, line 5: Can we have this occur when we push a new `xX.Y` release tag? We'd like the GH pages to reflect the latest release, rather than `main`

nahoc: I updated the action to make it push on tags 👍

nategraf:
So, the links in the README lead to 404. This is not specific to the work, but if you have time to look into solutions there, it would be appreciated. Maybe we should just link back to GitHub. Maybe we should have something like what `clap` does for their docs https://docs.rs/clap/latest/clap/_derive/index.html

nahoc:
> So, the links in the README lead to 404. This is not specific to the work, but if you have time to look into solutions there, it would be appreciated. Maybe we should just link back to GitHub. Maybe we should have something like what `clap` does for their docs https://docs.rs/clap/latest/clap/_derive/index.html

Right, so the easiest way to make this work atm is just to use absolute links instead of relative in markdown.

Otherwise I'll try digging a bit to see if I can find something.

nategraf (approved):

PR #420, opened 2025-01-27
WEB3-322: feat: Add `HostEvmEnv::extend` method for merging environments

Wollac:
This PR adds the ability to merge multiple `HostEvmEnv` instances, enabling parallel preflight execution of contract calls.

## Changes

- Add `extend` method to `HostEvmEnv` for merging environments
- Add comprehensive documentation with examples
- Add `extend_checked` helper function for safe HashMap merging

## Usage Example
```rust
// Create separate environments for parallel execution
let mut env1 = builder.clone().build().await?;
let mut contract1 = Contract::preflight(usdt_addr, &mut env1);
let mut env2 = builder.clone().build().await?;
let mut contract2 = Contract::preflight(usdc_addr, &mut env2);
// Execute calls in parallel
tokio::join!(
  contract1.call_builder(&call).call(),
  contract2.call_builder(&call).call()
);
// Merge the environments
env1.extend(env2);
let evm_input = env1.into_input().await?;
```

## Implementation Details

- Merges database state
- Performs consistency checks to ensure environments are compatible
- Panics on inconsistencies (mismatching configs, headers, or DB states)
- Preserves all proofs and state from both environments

## Testing

The functionality is demonstrated through doctest examples. Additional integration tests would be valuable to add in a follow-up PR.

### DIFF
--- a/crates/steel/src/host/db/proof.rs
+++ b/crates/steel/src/host/db/proof.rs
@@ -92,7 +92,7 @@ impl<D> ProofDb<D> {
 
     /// Extends the `ProofDb` with the contents of another `ProofDb`. It panics if they are not
     /// consistent.
-    pub fn extend(&mut self, other: ProofDb<D>) {
+    pub(crate) fn extend(&mut self, other: ProofDb<D>) {
         extend_checked(&mut self.accounts, other.accounts);
         extend_checked(&mut self.contracts, other.contracts);
         extend_checked(&mut self.proofs, other.proofs);
--- a/crates/steel/src/host/db/proof.rs
+++ b/crates/steel/src/host/db/proof.rs
@@ -58,9 +58,12 @@ struct StorageProof {
     proof: Vec<Bytes>,
 }
 
-impl<D: Database> ProofDb<D> {
+impl<D> ProofDb<D> {
     /// Creates a new ProofDb instance, with a [Database].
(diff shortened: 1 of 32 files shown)


crates/steel/src/host/db/proof.rs
```diff
@@ -83,6 +89,15 @@ impl<D: Database> ProofDb<D> {
     pub fn inner(&self) -> &D {
         &self.inner
     }
+
+    /// Extends the `ProofDb` with the contents of another `ProofDb`. It panics if they are not
+    /// consistent.
```
nategraf, line 94: A `ProofDb` is linked to a single state (commitment) ya? So trying to merge the `ProofDb` for two different blocks will panic here. I can see how, if written correctly, this will never fail. It also seems like an common enough mistake that panicking is perhaps too extreme, and could result in DoS in cases where e.g. someone is running a service that produces Steel proofs, if an error can cause the service to go down.

Wollac: I changed it so that the public extend now returns an error if the header config does not match. This could be a user error.
If the actual hash maps are extended, the panic is kept because this should not be possible without fraudulent RPC nodes. Also, this would be unrecoverable since the DB has already been modified.

nategraf (approved):

PR #421, opened 2025-01-27
WEB3-323: fix: use rzup

Wollac:
Installs the rust toolchain using rzup.

this fixes the following error we are currently seeing in the CI:
```
cargo risczero install --version r0.1.81.0
Error: Run `rzup install` instead
```

This PR also contains #424  which is necessary to make the tests pass.

### DIFF
--- a/.github/actions/cargo-risczero-install/action.yaml
+++ b/.github/actions/cargo-risczero-install/action.yaml
@@ -61,9 +61,11 @@ runs:
       run: cargo install --locked --bin r0vm --path risc0/cargo-risczero --features "${{ inputs.features }}"
       shell: bash
       working-directory: tmp/risc0
-    - name: install rust toolchain
+    - name: install toolchains
       if: inputs.ref != ''
-      run: cargo run --bin rzup -- install --verbose --force rust ${{ inputs.toolchain-version }}
+      run: |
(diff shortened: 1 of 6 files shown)


.github/actions/cargo-risczero-install/action.yaml
```diff
@@ -30,47 +30,49 @@ inputs:
...
+      shell: bash
 
-      - name: install toolchains
-        run: cargo risczero install ${{ inputs.toolchain-version != '' && format('--version {0}', inputs.toolchain-version) || '' }}
-        shell: bash
-      - name: cleanup
-        run: rm -rf tmp
-        shell: bash
+    - name: install rust toolchains
+      run: cargo run --bin rzup -- --verbose install rust ${{ inputs.toolchain-version }}
```
nategraf, line 72: `cargo run --bin rzup` will only work when `ref != ''` (i.e. when a ref is specified and not a version). When a version is specified, I'm not sure what the best approach is. Ideally, we could install `rzup` via curl-to-bash or from the latest release on `crates.io` and run `rzup install ${{ version }}` but that is not actually a valid command and I'm not immediately sure what the long-form version of this is.

capossele (approved):

nategraf (approved):

PR #424, opened 2025-01-28
WEB3-328: fix: Use a builder pattern for GuestOptions and DockerOptions

Wollac:
https://github.com/risc0/risc0/pull/2762 changed the way GuestOptions are constructed. This PR fixes this and switches to the new builder pattern.


nategraf (approved):

nategraf:
I'm going to skip CI checks here, since they are broken and waiting for https://github.com/risc0/risc0-ethereum/pull/421, which is independently required to fix CI.

PR #425, opened 2025-01-28
WEB3-329: chore: Add op-steel readme

Wollac:


crates/op-steel/README.md
nategraf, line 1: This README does a good job of discussing mechanics. It may not be immediately clear to the reader they can use this for cross-chain Steel queries / cross-chain communication. It might be good to highlight that, as I think its the most common use case for this work.

nategraf (approved):

PR #432, opened 2025-01-31
chore: Enable keccak precompile in examples

Wollac:


nategraf (approved):
Does this improve performance in the case of making just a few SLOADs?

Wollac:
> Does this improve performance in the case of making just a few SLOADs?

I did some very rough benchmarks on this and at least for the smallest realistic use cases (like loading a single account from Sepolia) it is still a significant speedup (almost 3x). I haven't benchmarked artificial cases like an Anvil with only a single account with empty state or anything like that, but this seems to be enough to add it to the examples and recommend it.

PR #454, opened 2025-02-22
WEB3-359: Make compatible with v2

capossele:


crates/ffi/tests/basic.rs
```diff
@@ -44,7 +44,7 @@ fn basic_usage() {
     assert_eq!(journal, hex::decode("deadbeef").unwrap());
     let expected_receipt = Receipt::new(
         InnerReceipt::Fake(FakeReceipt::new(ReceiptClaim::ok(
-            ECHO_ID,
+            compute_image_id_v2(Digest::from(ECHO_ID)).unwrap(),
```
flaub, line 47: You can just say `ECHO_ID.into()`

rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.81"
```
Wollac, line 2: What is the motivation for this change? In general, this should match the rust version of the zkVM which is still at 1.81

capossele: I was just following what used in risc0. Yeah it is odd though that does not match with what supported by the zkvm

Wollac (approved):

PR #456, opened 2025-02-27
WEB3-361: chore: Remove `munge` dependency

Wollac:
The recently released 0.4.3 of `munge` fixes issues with the previous 0.4.2, so the additional locked dependency is no longer needed.


capossele (approved):
Oh that’s good, finally!

PR #460, opened 2025-03-03
Fix method naming inconsistency in EvmEnvBuilder

VolodymyrBg:
Rename commitment_block_number method to commitment_block_number_only in EvmEnvBuilder class to maintain consistent naming convention with other similar methods like commitment_block_hash and commitment_block_number_or_tag. This change improves code readability and makes the API more intuitive.


Wollac:
This was introduced to be consistent with the corresponding block_hash, etc. methods of the builder.

PR #463, opened 2025-03-05
WEB3-367: Install r0vm

capossele:
In the cargo-risczero-install action, specifically the path using the version (rather than the ref) the r0vm installation was missing

### DIFF
--- a/.github/actions/cargo-risczero-install/action.yaml
+++ b/.github/actions/cargo-risczero-install/action.yaml
@@ -75,13 +75,11 @@ runs:
       working-directory: tmp/risc0
 
     # Only run this step if using version (installing from crates.io)
-    - name: install cargo-risczero from crates.io
+    - name: install rzup
       if: inputs.version != ''
-      run: cargo install --locked cargo-risczero --version "${{ inputs.version }}" --no-default-features --features "${{ inputs.features }}"
-      shell: bash
-    - name: install rzup from git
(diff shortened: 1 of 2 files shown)


rust-toolchain.toml
```diff
@@ -1,5 +1,5 @@
 [toolchain]
-channel = "1.81"
+channel = "1.83"
```
Wollac, line 2: I am really not a fan of this, as - especially for risc0-ethereum - it leads to super cryptic error messages, as alloy has already switched to rust 1.83, but all these crates are not (yet) supported in our ZKVM.

Wollac (approved):
To make the toolchain version of `risc0-ethereum` independent of the toolchain version of `risc0`, I added the use of `rustup` to ensure that the active version is installed.
I also changed it to download `rzup` instead of building it from scratch.

.github/actions/cargo-risczero-install/action.yaml
```diff
@@ -73,10 +78,15 @@ runs:
     - name: install cargo-risczero from crates.io
```
Wollac, line 78: Do we actually still need to install cargo-risczero at all? Hasn't it been completely replaced by rzup?

rust-toolchain.toml
Wollac: Has been addressed in ab9731892614ea0a4afc82ad82f878c224a13255

PR #477, opened 2025-03-10
WEB3-374: chore: Update alloy to v0.12

Wollac:
Blocked until rust 1.85 is supported by the guest.


Wollac:
The PR can be re-opened when we support rust 1.85 in the zkVM to also update op-alloy

capossele (approved):

PR #502, opened 2025-03-20
WEB3-385: Update the ControlID for latest main of risc0

bobbobbio:


bobbobbio:
already done in https://github.com/risc0/risc0-ethereum/pull/501

bobbobbio:
I was wrong, seems like there was another change to these? I ran bootstrap and I had the commits from #501 already actually

capossele:
#501 should suffices, the changes here are only related to the test receipt that changes at every bootstrap

bobbobbio:
got it, thanks!

PR #506, opened 2025-03-25
WEB3-386: feat: consensus commitment

Wollac:
This PR introduces a new commitment version the "Consensus Commitment" which is a Steel commitment that can be validated against the beacon chain state directly.


crates/steel/src/host/builder.rs
```diff
@@ -224,11 +234,39 @@ impl<P, H> EvmEnvBuilder<P, H, ()> {
...
 pub struct History {
-    beacon_url: Url,
+    beacon_config: Beacon,
     commitment_block: BlockId,
 }
 
-impl<P> EvmEnvBuilder<P, EthBlockHeader, Url> {
+impl<P> EvmEnvBuilder<P, EthBlockHeader, Beacon> {
+    /// Creates a consensus commitment instead of a beacon commitment that can be verified using the
+    /// EIP-4788 beacon root contract.
```
nategraf, line 243: This sentence is a bit unclear which of the two, consensus or beacon, can be verified using the EIP-4788 beacon root contract. What about the following?
```suggestion
    /// Creates a consensus commitment, with a slot number as the identifier.
```

nategraf (approved):
Even though SteelVerifier.sol cannot handle version 2, can you add a check in that contract for when it is passed and return a specific error? It would be better for devex and UX to get something like "stateless SteelVerifier does support consensus slot commitments" (or something). Also, with 3-4 versions in use now I would like to have a list somewhere of what each version number is assigned to. The smart contract seems like a reasonable place, but anywhere else would do too.

crates/steel/src/host/builder.rs
Wollac: addressed in d2f9b9a3615b9d0adcd42a8eb537e759db6321a2

PR #534, opened 2025-04-02
minor docstring fixups

strmfos:
There are four hyphens at the beginning of a PGP block instead of five. The correct form is: `-----BEGIN PGP PUBLIC KEY BLOCK-----`


Wollac (approved):
That is technically correct

PR #561, opened 2025-05-05
WEB3-433: fix: Add configID verification to SteelVerifier

Wollac:


crates/steel/src/host/mod.rs
```diff
@@ -156,6 +156,15 @@ pub struct HostCommit<C> {
     config_id: B256,
 }
 
+impl<C> HostCommit<C> {
+    /// Returns the config ID.
+    #[allow(dead_code)]
```
nategraf, line 161: Why is this `#[allow(dead_code)]` here? If this is only used under some configurations, then we could use `cfg`

nategraf (approved):
Looks good 👍

crates/steel/src/host/mod.rs
Wollac: Fixed in 21ff9c54dd4391c1a5dc0df4c06ae0cd08fa9e8e

PR #585, opened 2025-05-13
docs: replace outdated Foundry scripting link

eeemmmmmm:
The old tutorial URL returned 404. Updated it to the new canonical page:
https://book.getfoundry.sh/guides/scripting-with-solidity


nategraf:
It does appear that link is dead. The replacement (`https://book.getfoundry.sh/guides/solidity-scripting`) is also a dead link though.

nategraf (approved):

PR #608, opened 2025-06-02
WEB3-462: Migrate CI to cluster

flaub:


nategraf (approved):

nategraf:
The failure in https://github.com/risc0/risc0-ethereum/actions/runs/15401926036/job/43336360076?pr=608 is expected right now, until https://github.com/risc0/risc0-ethereum/pull/606 lands. As far as testing that CI works, that's a ✅

PR #619, opened 2025-06-10
Revert "WEB3-462: Migrate CI to cluster" / Pin Actions

morgangibbs87:
- Reverts risc0/risc0-ethereum#608
- Pins actions to proper commit shas

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,7 +55,7 @@ jobs:
           fetch-depth: 0
       - uses: risc0/risc0/.github/actions/rustup@fbe1d0bb75c21fe36cefd87bae25f424b711b291
       - name: Install cargo-sort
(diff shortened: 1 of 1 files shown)


capossele:
I think prod uses an older Ubuntu, as such the latest compatible version of Forge with that is 1.0.0, so maybe you should try changing that as well

.github/workflows/e2e.yml
```diff
@@ -34,15 +34,14 @@ jobs:
...
-            nvcc_arch: sm_89
+            device: nvidia_rtx_a5000
         # Exclude Bonsai proving on non-release branches.
         exclude:
           - release: false
             prover: bonsai
     env:
-      NVCC_APPEND_FLAGS: -arch=${{ matrix.nvcc_arch }}
       RUST_BACKTRACE: full
+      SCCACHE_RECACHE: 1
```
flaub, line 44: We should drop this

PR #623, opened 2025-06-18
Add null address validation for verifiers

aso20455:
## Changes
- Add `VerifierAddressZero` error to both contracts
- Add null address check in `RiscZeroVerifierRouter.addVerifier()`
- Add null address check in `RiscZeroSetVerifier` constructor
- Add unit tests for both validation cases
- Fix copyright year in affected files (2025 → 2024)

## Security considerations
This change improves contract security by preventing operations with null address verifiers that would inevitably fail at runtime, potentially causing unexpected behavior in dependent systems.


nategraf:
This seems like a reasonable way to create a non-functional verifier by accident.

Could you please verify your commit by signing it? Only signed commits are allowed in releases.

nategraf (approved):

PR #629, opened 2025-06-27
replace dead reference steel.rs

reject-i:
Hi devs, I found a broken link and decided to replace it, but I see there are many versions with identical file content. If you can suggest a more correct replacement link, I’d be glad to help. Thank you


nategraf:
The new link to not equivalent. It points to a different test case.

Wollac (approved):

Wollac:
Will be merged, once https://github.com/risc0/risc0-ethereum/pull/659 is in

PR #643, opened 2025-07-17
WEB3-499: chore: upgrade `alloy-evm`

Wollac:
- Update all alloy crates to the latest version
- Update Rust toolchain to v1.88 (Rust 1.86 is now required by alloy)
- Fix warnings for updated clippy


crates/op-steel/src/optimism/mod.rs
```diff
@@ -171,17 +173,37 @@ impl EvmBlockHeader for OpBlockHeader {
...
+                    ),
+                    SpecId::OSAKA => BlobExcessGasAndPrice::new(
+                        excess_blob_gas,
+                        eip7691::BLOB_GASPRICE_UPDATE_FRACTION_PECTRA as u64,
+                    ),
+                    _ => unimplemented!(
+                        "unsupported spec with `excess_blob_gas`: {}",
+                        <&'static str>::from(spec)
+                    ),
+                });
```
nategraf, line 196: I ran into this with a bit of my work. Can we have a public utility function for this, or do you think that would be a bad idea?

nategraf (approved):

PR #644, opened 2025-07-17
fix(op-steel): Replace unwrap() with proper error handling in l1.rs

rnkrtt:
Replace block_response.unwrap() with block_response.ok_or(Error::NotYetPropagated)? to properly handle the case when a block is not found


Wollac:
merged as https://github.com/boundless-xyz/steel/pull/3

PR #664, opened 2025-08-11
Bootstrap control IDs and test receipt from risc0-zvkm v3.0.0-rc.3

nategraf:


capossele (approved):

.github/workflows/e2e.yml
```diff
@@ -62,9 +62,9 @@ jobs:
         env:
           DOCKERHUB_USER: ${{ secrets.DOCKERHUB_CI_USER }}
       - if: matrix.feature == 'cuda'
-        uses: risc0/risc0/.github/actions/cuda@fbe1d0bb75c21fe36cefd87bae25f424b711b291
-      - uses: risc0/risc0/.github/actions/rustup@fbe1d0bb75c21fe36cefd87bae25f424b711b291
-      - uses: risc0/risc0/.github/actions/sccache@fbe1d0bb75c21fe36cefd87bae25f424b711b291
+        uses: risc0/risc0/.github/actions/cuda@main
+      - uses: risc0/risc0/.github/actions/rustup@main
+      - uses: risc0/risc0/.github/actions/sccache@main
```
flaub, line 67: We should not do this, we should pin to a specific version so that builds don't break randomly in the future.

PR #668, opened 2025-08-14
WEB3-530: Remove Steel for this repository

nategraf:
Steel is now maintained in its own repository. With this, it no longer needs to be in this repo, and users should reference that repository instead. https://github.com/boundless-xyz/steel


examples/governance/README.md
Wollac, line 1: I don't think the governance example contains anything related to steel. Therefore, it has not been included in Boundless/Steel.
Therefore, we should consider whether and where we want to keep it.

nategraf: Its super old and not representative of the best patterns anymore. Hence my deleting it here

PR #700, opened 2025-10-20
WEB3-548: Add RiscZeroVerifierLayeredRouter

capossele:
This PR proposes the introduction of a new smart contract, `RiscZeroVerifierLayeredRouter`, which extends the functionality of the existing `RiscZeroVerifierRouter` by incorporating a fallback mechanism to a canonical verifier router. This allows for the secure addition of potentially higher-risk verifiers (e.g., third-party implementations) while reusing verifiers from a trusted canonical source.

### DIFF
--- a/contracts/test/RiscZeroVerifierLayeredRouter.t.sol
+++ b/contracts/test/RiscZeroVerifierLayeredRouter.t.sol
@@ -24,7 +24,6 @@ import {
     IRiscZeroVerifier,
     Output,
     OutputLib,
-
     // Receipt needs to be renamed due to collision with type on the Test contract.
     Receipt as RiscZeroReceipt,
     ReceiptClaim,


contracts/src/RiscZeroVerifierFallbackRouter.sol
```diff
@@ -0,0 +1,78 @@
...
+
+import {IRiscZeroVerifier} from "./IRiscZeroVerifier.sol";
+import {RiscZeroVerifierRouter} from "./RiscZeroVerifierRouter.sol";
+
+/// @notice Router for IRiscZeroVerifier, allowing multiple implementations to be accessible behind a single address
+///         and a fallback to the canonical router.
+/// @dev Extends RiscZeroVerifierRouter to add fallback behavior.
+contract RiscZeroVerifierFallbackRouter is RiscZeroVerifierRouter {
+    /// @notice The canonical RISC Zero verifier router used as fallback.
+    IRiscZeroVerifier public immutable fallbackRouter;
```
nategraf: To function properly, this needs to be `RiscZeroVerifierRouter`. Instead of casting it in `addVerifier`, we could simply change the type here.
```suggestion
    RiscZeroVerifierRouter public immutable fallbackRouter;
```

capossele:
> Looks good. I do have a few comments.
> 
> On the name "fallback" doesn't quite give the right idea, in that the upstream router will be used most of the time. The best name I could think of would be something like `OverrideRouter` where the RISC Zero router would be the `base` or `parent`.

I've suggested LayeredRouter and called the canonical RISC Zero `parentRouter`

capossele:
Moving this over to boundless

